PR #8, opened 2020-08-24
ON-7471 # Added useBooleanState doc

Zaxist:


docs/useBooleanState.md
```diff
@@ -0,0 +1,33 @@
...
+As such, the items in the array can be destructured and named whatever you like:
+
+```js
+import { useBooleanState } from '@oneblink/apps-react'
+
+const [
+  isOpen /* true */,
+  setIsOpen,
+  setIsNotOpen,
+  ToggleIsOpen,
```
mymattcarroll, line 31: Thoughts on changing the example variables: `isDialogOpen`, `openDialog`, closeClose`, `toggleDialog`. Just a but clearer IMO.

Zaxist: yep

PR #9, opened 2020-08-24
ON-7470 # Added nullableState doc

Zaxist:


docs/useNullableState.md
```diff
@@ -0,0 +1,51 @@
...
+| Property       | Type | Description                      |
+| -------------- | ---- | -------------------------------- |
+| `defaultValue` | `T`  | The starting state for the hook. |
+
+## Return
+
+The return type of `useNullableState(myState)` is an array where:
+
+- The first item is of type `T` (the state).
+- The second item is `(newValue: T) => void` (a function that sets the state to the `newValue`).
```
mymattcarroll, line 20: There is a third element in the array which is a function to clear the state (set to `null`)

Zaxist: ahh, so there is.

docs/useNullableState.md
```diff
@@ -0,0 +1,59 @@
...
+} */,
+  setUserProfile /* A function that takes an object typed as {
+    name: string,
+    profession: string
+  } */,
+  unsetUserProfile,
+] = useBooleanState(startingProfile)
+```
+
+`setuserProfile` can then called with an object of type `T` like:
```
mymattcarroll, line 46:
```suggestion
`setUserProfile` can then called with an object of type `T` like:
```

docs/useNullableState.md
```diff
@@ -0,0 +1,59 @@
+# OneBlink Apps-react | useNullableState Hook
+
+## `useNullableState()`
+
+This function is a react hook for state of type of your choosing. It comes with a `useCallback` for setting the state.
```
mymattcarroll, line 5:
```suggestion
This function is a react hook for state of type of your choosing. It comes with two memoized functions, one for setting state and one for clearing it.
```

PR #10, opened 2020-08-24
ON-7472 # Added isOffline doc

Zaxist:


docs/useIsOffline.md
```diff
@@ -0,0 +1,17 @@
...
+## Return
+
+The return type of `useIsOffline()` is a `boolean`
+
+The hook can be used like:
+
+```js
+import { useIsOffline } from '@oneblink/apps-react'
+
+const isOffline = useBooleanState()
```
mymattcarroll, line 16:
```suggestion
const isOffline = useIsOffline()
```

docs/useIsOffline.md
```diff
@@ -0,0 +1,17 @@
+# OneBlink Apps-react | useIsOffline Hook
+
+## `useIsOffline()`
+
+This function is a react hook for determining whether an application is in an offline state.
```
mymattcarroll, line 5: Will need to also document the `IsOfflineContextProvider`. I think it makes sense to have those docs in this file as well. Thoughts?

Zaxist: yeh probably makes sense

PR #13, opened 2020-08-24
ON-7455 # Added docs for OneBlinkForm

mymattcarroll:


docs/OneBlinkForm.md
```diff
@@ -0,0 +1,274 @@
+# OneBlink Apps - ReactJS | Usage
+
+## `<OneBlinkForm />`
+
+Component for rendering a OneBlink Form. This component will render the submit, cancel and save draft buttons but it is up to the developer to implement what happens when those buttons are clicked.
+
+It is also recommended to import the `(s)css` for from this library as well.
```
Zaxist, line 7:
```suggestion
It is also recommended to import the `(s)css` from this library as well.
```

docs/OneBlinkForm.md
```diff
@@ -0,0 +1,274 @@
...
+        if (isMounted.current) {
+          setPostSubmissionActionErrorMessage(error.message)
+        }
+      }
+    },
+    [history.push, isMounted],
+  )
+
+  const handleSubmit = React.useCallback(
+    async (formSubmission: FormSubmission) => {
```
Zaxist, line 120: These typings won't be valid to anyone not using our flow types (no one?). We could `import { FormTypes } from '@oneblink/apps' and then use `FormTypes.FormSubmission` to make the code valid ts?

mymattcarroll: Sounds good.

PR #24, opened 2020-10-28
ON-8715 # Disabled further submission after submitting form

Zaxist:


src/OneBlinkForm.js
```diff
@@ -294,7 +295,7 @@ function OneBlinkForm(
   const handleSubmit = React.useCallback(
     (event) => {
       event.preventDefault()
-
+      if (hasSubmitted) return
```
mymattcarroll, line 298: Unfortunately, we can't do this as the onSubmit may not succeed, and we are not waiting for a successful response. E.g. if the request to submit fails the user should be able to press submit and try again. This would prevent that.

PR #25, opened 2020-10-28
ON-8742 # Added disabled prop to OneBlinkForm component

Zaxist:


mymattcarroll:
Need to update the docs as well.

mymattcarroll:
I think we probably need to disable the Cancel and Save Draft buttons as well, based on this prop.

mymattcarroll:
Are there any updates required to the AutoSaveOneBlinkForm component?

Zaxist:
> Are there any updates required to the AutoSaveOneBlinkForm component?

No

PR #35, opened 2020-11-09
ON-8895 # Changed services to typescript

aaronroworth:


src/services/barcode-readers/quagger.ts
```diff
@@ -17,10 +14,7 @@ const BARCODE_TYPES = [
   'code_93_reader',
 ]
 
-export default function (
-  imgData /* : string */,
-  barcodeTypes /* : string[] | void */,
-) /* : Promise<string | void> */ {
+export default function (imgData: string, barcodeTypes: string[] | void) {
```
Zaxist, line 17:
```suggestion
export default function (imgData: string, barcodeTypes?: string[]) {
```
Can we make this optional?
or else `undefined`? as `void` behaves differently to `undefined` in ts as oppposed to flow (no optional chaining on void, can't pass to other optional arg fns)
```suggestion
export default function (imgData: string, barcodeTypes: string[] | undefined) {
```

src/services/defaultCoordinates.ts
```diff
@@ -8,6 +6,7 @@ const auCoords = {
   longitude: 151.20863228242192,
 } //CIRCULAR QUAY, SYDNEY
 const coords = () =>
+// @ts-ignore
```
Zaxist, line 9: This is fixed in the latest `apps` beta

src/services/form-validation.ts
```diff
@@ -0,0 +1,545 @@
...
+    },
+    {},
+  )
+
+  return validate.isEmpty(pagesValidation) ? undefined : pagesValidation
+}
+
+const clearValidationMessagesForHiddenElements = (
+  formElementsValidation: FormElementsValidation | void,
+  formElementsConditionallyShown: FormElementsConditionallyShown | void,
```
Zaxist, line 153: Can these be optional or `undefined` too?

PR #42, opened 2020-11-20
ON-9175 # added user token to request header

simonmarklar:


src/components/LookupNotification.tsx
```diff
@@ -348,6 +349,12 @@ async function fetchLookup(
     }
   }
 
+  // check for userToken in the query string
+  const qs = querystring.parse(location.search)
+  if (qs.userToken) {
+    headers['X-OneBlink-User-Token'] = qs.userToken as unknown as string
+  }
```
mymattcarroll, line 356: Ideally this comes from the authService as well like the Id token. That way devs can use it without having to know what query string parameters are required to use it.

simonmarklar: yeah makes sense but auth service comes from apps which has not been updated yet

src/components/LookupNotification.tsx
```diff
@@ -348,6 +348,12 @@ async function fetchLookup(
     }
   }
 
+  // check for userToken in the query string
```
mymattcarroll, line 351: comment is outdated.

src/form-elements/FormElementAutocomplete.tsx
```diff
@@ -1,6 +1,7 @@
 import * as React from 'react'
 import AbortController from 'abort-controller'
 import clsx from 'clsx'
+import querystring from 'query-string'
```
mymattcarroll, line 4: can we remove this?

src/form-elements/FormElementAutocomplete.tsx
```diff
@@ -162,6 +163,12 @@ const AutocompleteFetch = React.memo(function AutocompleteFetch({
           }
         }
 
+        // check for userToken in the query string
```
mymattcarroll, line 166: comment is outdated.

PR #53, opened 2020-12-20
ON-9609 # Added contrast text determination to radio and checkbox but…

Zaxist:
…tons


src/form-elements/FormElementRadio.tsx
```diff
@@ -122,4 +107,41 @@ function FormElementRadio({
   )
 }
 
+type RadioOptionButtonProps = {
+  element: Props['element']
+  option: FormTypes.ChoiceElementOption
+  isSelected: boolean
+  onClick: () => void
+}
+const RadioOptionButton = React.memo(function RadioOptionButton({
```
simonmarklar, line 116: the only difference i can see between this and `CheckboxOptionButton ` is the class names, is it worth making this able to accept those as props? 

eg 

```
{filteredOptions.map((option) => (  <OptionsButton {...props} className="ob-button-radio-container" buttonClassName="ob-radio__button cypress-radio-button-control" /> )
```

Zaxist: yep sounds like a good idea

PR #54, opened 2021-01-13
ON-9681 # Added geoscapeAddress form element

mymattcarroll:


src/components/AutocompleteDropdown.tsx
```diff
@@ -0,0 +1,327 @@
...
+  id,
+  label,
+  value,
+  placeholder,
+  required,
+  disabled,
+  validationMessage,
+  displayValidationMessage,
+  searchDebounceMs,
+  searchMaxCharacters,
```
aaronroworth, line 42: Not a big deal, but should this prop be called `searchMinCharacters` rather than `searchMaxCharacters`? Max kinda implies its the most characters you can search with?

mymattcarroll: agreed

PR #64, opened 2021-02-12
ON-10125 # Change order of imports for bulma variables and styles

mymattcarroll:


src/styles.scss
```diff
@@ -27,45 +17,50 @@ $elevation-3: 0 1px 8px 0 rgba(0, 0, 0, 0.2), 0 3px 4px 0 rgba(0, 0, 0, 0.14),
...
+// Our Variables
+//
+$mobile: $tablet - 2px;
+$secondary: $grey-lighter;
 
 //
 // Bulma Styles
 //
-@import 'bulma/sass/utilities/_all';
+@import 'bulma/sass/base/_all';
```
simonmarklar, line 34: one of the previous tickets was to remove this, right? i take it the re-ordering solves the problem and thats why base is back ?

mymattcarroll: Leaving this here as removing it makes a bunch of things just look stupid. i.e. anyone that uses this lib will need to include it themselves. Going to handle the use case in the Forms CDN lib using this method:

```scss
.oneblink-apps-react-styles {
  box-sizing: border-box;

  @import '~@oneblink/apps-react/dist/styles.scss';
}
```

As the only styles added to `html` and `body` are most likely already going to be set in the consuming application.

simonmarklar: sweet, makes sense :)

PR #70, opened 2021-02-17
ON-10232 # Handle crash with non image files and watermark included

kizaonline:


src/form-elements/FormElementCamera.tsx
```diff
@@ -88,43 +88,44 @@ function FormElementCamera({
...
+                      now,
+                      canvas.width -
+                        textWidth -
+                        backgroundPadding -
+                        backgroundMargin,
+                      canvas.height - 22,
+                    )
+                  }
+                }
                 base64data = canvas.toDataURL(file.type)
```
simonmarklar, line 126: wont this error because context will be null if `context = canvas.getContext('2d')` doesnt find one?

PR #72, opened 2021-02-24
ON-10316 # Added `useLoadDataState` hook

Zaxist:


docs/useLoadDataState.md
```diff
@@ -0,0 +1,64 @@
...
+| Property     | Type            | Description                                                                         |
+| ------------ | --------------- | ----------------------------------------------------------------------------------- |
+| `state`      | `LoadDataState` | The current state.                                                                  |
+| `handleLoad` | `function`      | A function that runs the `onLoad` function again. This can be used to refresh data. |
+
+`useLoadDataState()` can be used like:
+
+```js
+import { useLoadDataState } from '@oneblink/apps-react'
+import fetchData from 'fetchData' // Must be a function which returns a promise of some data
```
mymattcarroll, line 28: Can we change this to an example of a function that returns a promise with some state in it. Using `fetch` would be a good example.

Zaxist: yep

mymattcarroll:
```suggestion
const fetchData = async () => {
  const response = await fetch('https://example.com/data')
  return await response.json()
}
```

docs/useLoadDataState.md
```diff
@@ -0,0 +1,64 @@
...
+const [state, refresh] = useLoadDataState(fetchData)
+
+switch (state.status) {
+  case 'LOADING':
+    return <Loading />
+  case 'ERROR':
+    return <Error message={state.error} />
+  case 'SUCCESS':
+  // RENDER UI
+}
```
mymattcarroll, line 39: Can we wrap this part in a component function, it is a little bit misleading.

mymattcarroll:
Just need to add a link to the docs for the hooks here: `/docs/README.md`

PR #78, opened 2021-03-10
ON-10595 # Stopped lookups from triggering on read only forms

Zaxist:


src/components/LookupNotification.tsx
```diff
@@ -111,6 +113,8 @@ function LookupNotificationComponent({
   )
 
   const triggerLookup = React.useCallback(async () => {
+    // No lookups for read only forms
+    if (formIsReadOnly) return
```
mymattcarroll, line 117: `formIsReadOnly` should be added a dep to this callback too

PR #82, opened 2021-03-15
ON-10635 # Fixed calculations inside repeatable sets using values fro…

mymattcarroll:
…m all entries


src/form-elements/FormElementCalculation.tsx
```diff
@@ -241,32 +219,22 @@ function FormElementCalculation({
   // MODEL LISTENER
   React.useEffect(() => {
     if (!calculation) return
-    const submission = getFormSubmissionModel(false).submission
-    const newValue = calculation.eval(submission)
-    if (value === undefined) {
-      if (Object.is(NaN, newValue)) return
```
Zaxist, line 247: I have a memory that there was a  reason for using `Object.is` here..

mymattcarroll: `isNaN(newValue)` should be better suited here as `newValue` is not typed so it could be falsey.

PR #83, opened 2021-03-18
ON-10694 # Added compliance element

Zaxist:


src/components/OneBlinkFormElements.tsx
```diff
@@ -613,6 +614,31 @@ const FormElementSwitch = React.memo(function OneBlinkFormElement({
         </LookupNotification>
       )
     }
+    case 'compliance': {
+      return (
+        <LookupNotification
+          isAutoLookup
+          element={element}
+          elements={elements}
+          value={value}
```
mymattcarroll, line 623:
```suggestion
          value={value ? value.value : undefined}
```

src/form-elements/ComplianceButton.tsx
```diff
@@ -0,0 +1,35 @@
...
+  icon,
+  children,
+  isActive,
+  disabled,
+}: Props) => {
+  return (
+    <button
+      disabled={disabled}
+      type="button"
+      className={clsx('button compliance-button', {
```
mymattcarroll, line 23:
```suggestion
      className={clsx('button ob-button__compliance', {
```

src/form-elements/ComplianceButton.tsx
```diff
@@ -0,0 +1,35 @@
...
+        'is-primary': isActive,
+        'is-light': !isActive,
+      })}
+      onClick={onClick}
+    >
+      <i className="material-icons is-size-5 compliance-button-icon">{icon}</i>
```
mymattcarroll, line 29:
```suggestion
      <i className="material-icons is-size-5 ob-button__compliance-icon">{icon}</i>
```

src/form-elements/FormElementCompliance.tsx
```diff
@@ -0,0 +1,224 @@
...
+  onConditionallyShowOption: (
+    choiceElementOption: FormTypes.ChoiceElementOption,
+  ) => boolean
+  displayValidationMessage: boolean
+  validationMessage: string | undefined
+  isEven?: boolean
+}
+
+interface Value {
+  value: unknown
```
mymattcarroll, line 32:
```suggestion
  value: string
```

src/form-elements/FormElementCompliance.tsx
```diff
@@ -0,0 +1,224 @@
...
+  const notesElement = React.useMemo(() => generateNotesElement(element), [
+    element,
+  ])
+  const filesElement = React.useMemo(() => generateFilesElement(element), [
+    element,
+  ])
+
+  const handleValueChange = React.useCallback(
+    (fe: FormTypes.FormElement, v: unknown) => {
+      console.log({ v })
```
mymattcarroll, line 96: Can we update this to change some context or remove it.

src/form-elements/FormElementCompliance.tsx
```diff
@@ -0,0 +1,224 @@
...
+              onClick={toggleIsShowingFiles}
+              disabled={!typedValue.value}
+            >
+              Media
+            </ComplianceButton>
+          </div>
+          {isShowingNotes && (
+            <div className="ob-compliance-child-element">
+              <FormElementTextarea
+                id={`${element.id}-notes`}
```
mymattcarroll, line 198:
```suggestion
                id={`${id}-notes`}
```

This needs to be concatenated with the `id` property passed to this component. Otherwise the `id` will not be unique for entries in a repeatable set.

src/form-elements/FormElementCompliance.tsx
```diff
@@ -0,0 +1,224 @@
...
+                validationMessage={undefined}
+                value={typedValue.notes}
+                element={notesElement}
+              />
+            </div>
+          )}
+          {isShowingFiles && (
+            <div className="ob-compliance-child-element">
+              <FormElementFiles
+                id={`${element.id}-files`}
```
mymattcarroll, line 210:
```suggestion
                id={`${id}-files`}
```

src/styles/compliance.scss
```diff
@@ -0,0 +1,33 @@
+.ob-compliance {
```
mymattcarroll, line 1: ideally none of these are nested under this class. So that devs can override styles easier. Obviously these ones need to stay nested: `&.odd > .ob-compliance__container`

src/styles/compliance.scss
```diff
@@ -0,0 +1,33 @@
...
+  }
+
+  .compliance-button {
+    padding: 4px 16px;
+    height: unset;
+    font-size: $small-font-size;
+  }
+
+  .compliance-button-icon {
+    margin-right: 8px;
```
mymattcarroll, line 27: Can we change all these hard coded `px` and `rem` to `$size-*` variables. That way we will have consistent sizing throughout all components if those variables change.

src/form-elements/FormElementCompliance.tsx
Zaxist: This would require changing a lot more, including the hook that radeio buttons also use. Also, it can be an array as well

src/form-elements/FormElementCompliance.tsx
Zaxist: accident

src/form-elements/FormElementCompliance.tsx
mymattcarroll: it can be an array?

Zaxist: Sorry, no it can't, but the hook that is used for radios (and checkboxes) can spit the value out as an array, so it uses unknown

PR #85, opened 2021-03-18
ON-10750 # Added defaultValue for compliance

Zaxist:


src/services/generate-default-data.ts
```diff
@@ -14,7 +14,8 @@ export default function generateDefaultData(
       case 'checkboxes':
       case 'select':
       case 'autocomplete':
-      case 'radio': {
+      case 'radio':
+      case 'compliance': {
```
mymattcarroll, line 18: I dont think this will work. `compliance` element submission data is not the same as a radio button.

Zaxist: true

PR #102, opened 2021-04-12
ON-11235 # Removed delay to open hint tooltip

mymattcarroll:


src/components/FormElementLabelContainer.tsx
```diff
@@ -28,7 +28,7 @@ function FormElementLabelContainer({
           {element.label}
         </label>
         {element.hint && (
-          <Tooltip title={element.hint} arrow>
+          <Tooltip title={element.hint} arrow enterTouchDelay={0}>
```
Zaxist, line 31: Want to change the other tooltip(s ?) as well? Think theres one in the `BaseForm` component

mymattcarroll: Agreed

mymattcarroll: Actually, disagreed. It is on a button and a press on that should not show the tooltip, it should just execute the action. IMO

Zaxist: Sounds good

PR #103, opened 2021-04-12
ON-11231 # Increased hint font size

Zaxist:


src/styles/renderer.scss
```diff
@@ -234,3 +234,9 @@
 .ob-hint-tooltip {
   max-width: 400px;
 }
+
+.MuiTooltip-popper {
+  .MuiTooltip-tooltip {
```
mymattcarroll, line 239: I dont think these classes are applied when its built in production mode.

mymattcarroll: Nope, I was wrong.

PR #105, opened 2021-04-20
ON-11434 # Captured certain element failures with Sentry

Zaxist:


package.json
```diff
@@ -40,7 +40,7 @@
     "@babel/core": "^7.13.10",
     "@babel/preset-env": "^7.13.12",
     "@babel/preset-react": "^7.12.13",
-    "@oneblink/apps": "^0.7.2-beta.1",
+    "@oneblink/apps": "^0.7.3-beta.3",
```
mymattcarroll, line 43: Need to also update the `peerDependencies`

src/components/AutocompleteDropdown.tsx
```diff
@@ -193,6 +194,7 @@ function AutocompleteDropdown({
       try {
         newOptions = await onSearch(label, abortController.signal)
       } catch (error) {
+        Sentry.captureException(error)
```
mymattcarroll, line 197: Prob want to move this inside the `if` below. This error will throw to abort the http request which will happen if the user continues to type.

mymattcarroll:
I think we want to capture errors in the `useLogin()` hook and auto save data not loading/removing/updating in the `OneBlinkAutoSaveForm` component as well.

PR #113, opened 2021-04-23
ON-10187 # Added support for display as currency

Zaxist:


src/form-elements/FormElementCalculation.tsx
```diff
@@ -20,6 +20,7 @@ const isUnenteredValue = (value: unknown | undefined) => {
 }
 
 function FormElementCalculation({ element, onChange, value }: Props) {
+  element.displayAsCurrency = true
```
kizaonline, line 23: Was this for testing only?

kizaonline (approved):

package.json
```diff
@@ -40,7 +40,7 @@
     "@babel/core": "^7.13.10",
     "@babel/preset-env": "^7.13.12",
     "@babel/preset-react": "^7.12.13",
-    "@oneblink/apps": "^0.8.0-beta.1",
+    "@oneblink/apps": "^0.8.1-beta.1",
```
mymattcarroll, line 43: need to update `peer` as well.

PR #117, opened 2021-05-04
ON-11744 # Added onChange handling to nested submission data

mymattcarroll:


src/form-elements/FormElementCompliance.tsx
```diff
@@ -92,44 +81,89 @@ function FormElementCompliance({
...
+    [element, onChange],
   )
-  const handleFilesChange = React.useCallback(
-    (v: PossibleFileConfiguration[] | undefined) => {
-      const newFiles = v ? v : undefined
-      onChange(element, { ...typedValue, files: newFiles })
+  const handleFilesChange = React.useCallback<
+    React.ComponentProps<typeof FormElementFiles>['onChange']
+  >(
+    (fe, v) => {
```
Zaxist, line 131: Why do we need `fe` here? ie. why pass the element to the hook, only to pass it back here and seemingly do nothing with it?

mymattcarroll: `<FormElementFiles />` `onChange()` passes back the form element. If we leave it the way it is, the onChange prop will change every time this component renders.

PR #118, opened 2021-05-04
ON-11730 # Handled image preview

Zaxist:


src/form-elements/FormElementFiles/FormElementFileDisplay.tsx
```diff
@@ -1,26 +1,32 @@
...
-    if (file.type === 'NEW' || file.type === 'SAVING') {
-      return URL.createObjectURL(file.data)
+    if (attachment.type === 'NEW' || attachment.type === 'SAVING') {
+      return URL.createObjectURL(attachment.data)
     }
-    if (!file.type) {
-      // TODO: Load image
-      return
+    if (!attachment.type && attachmentBlob) {
+      return URL.createObjectURL(attachmentBlob)
```
mymattcarroll, line 27: These need to be cleaned up with https://developer.mozilla.org/en-US/docs/Web/API/URL/revokeObjectURL at some point

PR #122, opened 2021-05-05
ON-11788 # Handled download for files element

Zaxist:


src/hooks/attachments/useAttachment.ts
```diff
@@ -222,10 +222,21 @@ export default function useAttachment(
     }
   }, [value])
 
+  const canDownload = React.useMemo(() => {
+    return (
+      !!value &&
+      (typeof value === 'string' ||
+        value.type === 'NEW' ||
+        value.type === 'SAVING' ||
+        !!imageUrlState.imageUrl)
```
mymattcarroll, line 231: Should be able to download an attachment that is not an image.

src/services/download-file.ts
```diff
@@ -91,6 +91,34 @@ export default async function downloadFile(blob: Blob, fileName: string) {
...
+          },
+        }
+      : undefined,
+  )
+  if (!response.ok) {
+    throw new Error(
+      `Unable to download file. HTTP Status Code: ${response.status}`,
+    )
+  }
+  const blob = await response.blob()
```
mymattcarroll, line 118: All of this needs to be wrapped in that try catch. Downloading files should not throw an errors.

src/hooks/attachments/useAttachment.ts
```diff
@@ -228,9 +228,9 @@ export default function useAttachment(
       (typeof value === 'string' ||
         value.type === 'NEW' ||
         value.type === 'SAVING' ||
-        !!imageUrlState.imageUrl)
+        !value.type)
```
mymattcarroll, line 231: check is logged in and private/public as well.

src/services/download-file.ts
```diff
@@ -91,37 +116,24 @@ async function downloadFile(blob: Blob, fileName: string) {
...
-  }
-  const blob = await response.blob()
-  return await downloadFile(blob, attachment.fileName)
-}
-
 export async function downloadFileLegacy(dataURI: string, fileName: string) {
-  const blob = await convertDataUriToBlob(dataURI)
-  return await downloadFile(blob, fileName)
+  try {
+    const blob = await convertDataUriToBlob(dataURI)
```
mymattcarroll, line 121: This would be better moved into the cordova logic

src/services/download-file.ts
```diff
@@ -3,78 +3,103 @@ import * as bulmaToast from 'bulma-toast'
...
+  }
+}
+
+export default async function downloadAttachment(attachment: AttachmentValid) {
+  try {
+    if (attachment.type) {
+      return await downloadFile(attachment.data, attachment.fileName)
+    }
+    let idToken
+    if (attachment.isPrivate) {
```
mymattcarroll, line 83: if public, just use attachment.url.

src/hooks/attachments/useAttachment.ts
```diff
@@ -223,12 +223,14 @@ export default function useAttachment(
   }, [value])
 
   const canDownload = React.useMemo(() => {
+    const isDownloadableAttachment = (attachment: Attachment) =>
+      (attachment.type === 'NEW' ||
+        attachment.type === 'SAVING' ||
+        !attachment.type) &&
+      (!attachment.isPrivate || authService.isLoggedIn())
```
mymattcarroll, line 230: Should be using the `useAuth()` hook here. As it will trigger changes, `authService.isLoggedIn()` will only be evaluated once.

src/services/download-file.ts
```diff
@@ -92,8 +114,22 @@ export default async function downloadFile(blob: Blob, fileName: string) {
...
+      bulmaToast.toast({
+        message:
+          'Sorry, there was an issue downloading your file, please try again.',
+        // @ts-expect-error bulma sets this string as a class, so we are hacking in our own classes
+        type: 'ob-toast is-danger cypress-download-file-toast',
+        dismissible: true,
+        closeOnClick: true,
+      })
+    }
+  }
```
mymattcarroll, line 132: Can we please consolidate this duplicated error handling.

src/hooks/attachments/useAttachment.ts
```diff
@@ -222,10 +222,23 @@ export default function useAttachment(
     }
   }, [value])
 
+  const canDownload = React.useMemo(() => {
+    const isDownloadableAttachment = (attachment: Attachment) =>
+      (attachment.type === 'NEW' ||
+        attachment.type === 'SAVING' ||
+        !attachment.type) &&
```
mymattcarroll, line 229: I suppose there is no reason why we should not let an attachment that is in the ERROR state be downloaded.

src/hooks/attachments/useAttachment.ts
```diff
@@ -222,10 +224,20 @@ export default function useAttachment(
     }
   }, [value])
 
+  const canDownload = React.useMemo(() => {
+    const isDownloadableAttachment = (attachment: Attachment) =>
+      !attachment.isPrivate || isLoggedIn
```
mymattcarroll, line 229:
```suggestion
      !!attachment.type || !attachment.isPrivate || isLoggedIn
```

Should be allowing a private file to be downloaded if it has not been uploaded yet.

mymattcarroll: IMO, these types of memos are better suited to this format (favouring maintainability and readability over less code)

```js
if (firstPredicate) {
  return true
}

if (secondPredicate) {
  return true
}

if (thirdPredicate) {
  return true
}

return lastPredicate
```

Zaxist: Are you sure? I thought we wanted to deny downloading if it was private and you weren't logged in? Even it it hasnt uploaded yet. Doesn't make much sense to be inconsistent about that, even though there is a technical difference, user won't perceive it that way.

mymattcarroll: Agreed

PR #123, opened 2021-05-06
ON-11780 # Added compliance and files validation

Zaxist:


src/services/form-validation.ts
```diff
@@ -81,6 +80,38 @@ validate.validators.attachment = function (
...
+      const errorMessage = validateAttachment(att)
+      if (errorMessage) {
+        invalidAttachments.push(errorMessage)
+      }
+      return invalidAttachments
+    },
+    [],
+  )
+  if (invalidAttachmentErrors?.length) {
+    return `${invalidAttachmentErrors.join(', ')} could not be uploaded.`
```
mymattcarroll, line 97: If the error message for 2/3 files failing is: 

> Failed to fetch

The validation message returned will be  

> Failed to fetch, Failed to fetch could not be uploaded.

Is this intended?

Zaxist: No, I've just got the dumb

PR #127, opened 2021-05-06
ON-11781 # Corrected file orientation of `files` images

Zaxist:


src/services/parseFilesAsAttachments.ts
```diff
@@ -1,5 +1,36 @@
...
+  file: File | Blob,
+): Promise<HTMLCanvasElement> {
+  // @ts-expect-error For some reason, the types do not include this function returning a promise
+  const imageMetaData: loadImage.MetaData = await loadImage.parseMetaData(file)
+
+  console.log(
+    'Loading image onto canvas to correct orientation using image meta data',
+    imageMetaData,
+  )
+  const orientation = imageMetaData.exif?.get('Orientation')
```
mymattcarroll, line 15: I just learnt that if `orientation` is `undefined` or `1`, we can skip this step. Which should be true most of the time. Should be much faster if we don't have to load the image onto a canvas.

src/services/parseFilesAsAttachments.ts
```diff
@@ -77,6 +93,18 @@ export async function parseFilesAsAttachmentsLegacy(
   return attachments
 }
 
-export default function parseFilesAsAttachments() {
-  // TODO: Rotate blob files for new stoarge types
+export async function correctNewAttachmentOrientation(
```
mymattcarroll, line 96: Could you update the `camera` element to use this function as well.

Zaxist: Not worth doing until the annotation task imo

src/services/parseFilesAsAttachments.ts
```diff
@@ -77,6 +115,17 @@ export async function parseFilesAsAttachmentsLegacy(
...
+): Promise<AttachmentNew> {
+  try {
+    const blob = await correctFileOrientationAsBlob(attachment.data)
+    return {
+      ...attachment,
+      data: blob,
+    }
+  } catch (err) {
+    console.warn('Failed to rotate the orientation of a file:', err)
+    return attachment
```
mymattcarroll, line 129: If we are ignoring this error, prob want to send it to Sentry.

PR #159, opened 2021-06-03
ON-12498 # Added toggle all button

Zaxist:


src/form-elements/FormElementCheckBoxes.tsx
```diff
@@ -65,13 +73,23 @@ function FormElementCheckboxes({
...
+        {element.canToggleAll && (
+          <ToggleAllCheckbox
+            id={id}
+            element={element}
+            options={filteredOptions}
+            selected={selectedValues}
+            disabled={element.readOnly}
+            onChange={onChange}
+          />
+        )}
```
mymattcarroll, line 85: I think this should prob be inside `<FormElementOptions options={element.options}>` as this component will render a message/loading if there are no options and this should prob not be there if there are no options, or the options are still loading.

src/form-elements/FormElementSelect.tsx
```diff
@@ -45,6 +52,16 @@ function FormElementSelect({
...
+        {element.multi && element.canToggleAll && (
+          <ToggleAllCheckbox
+            id={id}
+            element={element}
+            options={filteredOptions}
+            selected={selectedValuesAsArray}
+            disabled={element.readOnly}
+            onChange={onChange as FormElementValueChangeHandler<string[]>}
+          />
+        )}
```
mymattcarroll, line 64: Same here

PR #160, opened 2021-06-07
ON-12670 # Added Bolean form element

Zaxist:


src/form-elements/FormElementBoolean.tsx
```diff
@@ -0,0 +1,65 @@
...
+          className={clsx('ob-boolean__input', 'cypress-boolean-control', {
+            'ob-boolean__input-checked': !!value,
+          })}
+          inputProps={{
+            className:
+              'PrivateSwitchBase-input-4 MuiSwitch-input cypress-boolean-input',
+          }}
+          value={!!value}
+          disabled={element.readOnly}
+          onBlur={setIsDirty}
```
mymattcarroll, line 48: Does a switch have a blur event?

Zaxist: Apparently, but no doesn't seem to work

src/components/FormElementLabelContainer.tsx
```diff
@@ -9,16 +9,19 @@ function FormElementLabelContainer({
...
   element: FormTypes.FormElementBase
   id: string
   required: boolean
   children: React.ReactNode
+  childrenFirst?: boolean
 }) {
   return (
     <div className={clsx('ob-form__element', className)}>
       <div className="label ob-label__container">
+        {childrenFirst && children}
```
mymattcarroll, line 24: Is this going to put the validation message(s) between the label and the switch?

Prob better to not use `<FormElementLabelContainer />` component at all. As i expect we dont really want the `*` on the label if its required.

src/form-elements/FormElementBoolean.tsx
```diff
@@ -0,0 +1,67 @@
...
+            })}
+            inputProps={{
+              className:
+                'PrivateSwitchBase-input-4 MuiSwitch-input cypress-boolean-input',
+            }}
+            value={!!value}
+            disabled={element.readOnly}
+            onChange={(e) => {
+              setIsDirty()
+              onChange(element, e.target.checked || undefined)
```
mymattcarroll, line 49: I dont think is right. This will prevent the PDF and such from showing "No". We should be setting `false` on the submission.

src/services/form-validation.ts
```diff
@@ -383,6 +389,21 @@ const generateSchemaReducer = (
         }
         break
       }
+      case 'boolean': {
+        if (formElement.type === 'boolean') {
```
mymattcarroll, line 393: Wont this always be true? Do we want to keep the log below?

src/form-elements/FormElementBoolean.tsx
Zaxist: sorry, yep changed this just before we spoke about it and forgot to change back

src/services/form-validation.ts
Zaxist: mistake, originally had this as part of a larger `case`

PR #191, opened 2021-06-24
ON-12997 # Added controlled versions of form components

mymattcarroll:


docs/OneBlinkFormControlled.md
```diff
@@ -0,0 +1,230 @@
...
+  OneBlinkForm,
+  useIsMounted,
+  useFormSubmissionState,
+} from '@oneblink/apps-react'
+import '@oneblink/apps-react/dist/styles.css'
+
+const captchaSiteKey = 'ENTER_YOUR_SITE_KEY_HERE'
+const googleMapsApiKey = 'ENTER_YOUR_MAPS_API_KEY_HERE'
+const formsAppId = 1
+const form: FormTypes.Form = {
```
Zaxist, line 42: Do we want to remove TS from JS example?

mymattcarroll: Nah, I wanted to leave them there. I think every project we are using this package in is now typescript.

docs/OneBlinkFormControlled.md
```diff
@@ -0,0 +1,230 @@
...
+  }
+
+  return (
+    <OneBlinkFormControlled
+      captchaSiteKey={captchaSiteKey}
+      googleMapsApiKey={googleMapsApiKey}
+      formsAppId={formsAppId}
+      definition={definition}
+      submission={submission}
+      setFormSubmission={setFormSubmission}
```
Zaxist, line 210: Should this just be `onSubmissionChange` ?

mymattcarroll: I wanted it to be clear that this is not just a callback type prop, but a react state setter.

PR #197, opened 2021-06-28
ON-13080 # Added extra validation for prefill data

Zaxist:


src/services/generate-default-data.ts
```diff
@@ -117,9 +131,95 @@ function parsePreFillData(
...
+      if (typeof value === 'number' && !isNaN(value)) {
+        return value
+      }
+      return
+    }
+    case 'boolean': {
+      if (typeof value === 'boolean') {
+        return value
+      }
+      return
```
mymattcarroll, line 164:
```suggestion
      return false
```

PR #235, opened 2021-11-19
ON-16953 # Added workaround for iOS15 date bug

Zaxist:


src/services/generate-default-data.ts
```diff
@@ -155,7 +155,9 @@ export function parseDateValue({
           day: '',
         },
       )
-    return `${year}-${month}-${day}`
+    // This is a workaround for an iOS 15 bug where the year was being returned as 2 digits: https://bugs.webkit.org/show_bug.cgi?id=230827
+    const fixedYear = year.length === 2 ? `20${year}` : year
```
mymattcarroll, line 159:
```suggestion
    const fixedYear = year.length === 2 ? date.getFullYear() : year
```

PR #236, opened 2021-11-22
ON-16987 # Added validation to check for allow extensionless attachments

jdawg093:


src/services/form-validation.ts
```diff
@@ -668,6 +668,15 @@ export function checkFileNameIsValid(
   )
 }
 
+export function checkFileNameHasExtension(
+  formElement: FormTypes.FilesElement,
+  fileName: string,
+) {
+  return (
+    formElement.allowExtensionlessAttachments || fileName.split('.').length <= 1
```
Zaxist, line 676: This should be 
```>= 1```
I think

File: src/services/form-validation.ts [hunk 2]
```diff
@@ -668,6 +668,15 @@ export function checkFileNameIsValid(
   )
 }
 
+export function checkFileNameHasExtension(
```
Zaxist, line 671: IMO we call this 
```checkFileNameExtensionIsValid```
since we are not explicitly checking if it has an extension, we are checking if the file name/extension is ok based on whether it has an extension and whether we allow extensions

PR #256, opened 2022-03-01
ON-19083 # Added trimming to user input elements

jdawg093:


Zaxist:
There is already an iteration over the submission happening when the user clicks submit. I think we should move this trim logic to there so we aren't iterating over it twice.

src/services/cleanFormSubmissionModel.ts
```diff
@@ -187,12 +187,20 @@ function cleanElementValue(
           break
         }
         if (stripBinaryData) {
+          const checklistObject = submission[element.name] as ComplianceValue
+          const value = checklistObject?.notes?.trim()
```
Zaxist, line 191: Should name this notes, IMO. `value` is ambiguous here

src/services/cleanFormSubmissionModel.ts
```diff
@@ -187,12 +187,20 @@ function cleanElementValue(
...
           model[element.name] = {
             ...(submission[element.name] as ComplianceValue),
```
Zaxist, line 193: This can be
```
...checklistObject
```

src/services/cleanFormSubmissionModel.ts
Zaxist, line 191: This stuff should be done before the `if` statement so we aren't repeating it in the `if` and `else`

src/services/cleanFormSubmissionModel.ts
```diff
@@ -210,6 +218,15 @@ function cleanElementValue(
...
       }
+      case 'text':
+      case 'textarea':
+      case 'email':
+      case 'barcodeScanner':
+      case 'telephone': {
+        const value = submission[element.name] as string
+        model[element.name] = value.trim()
+        break
+      }
```
Zaxist, line 229: This will prevent the conditional logic stuff below (in the 'default' clause) from running on these element types. We need to do this stuff inside that default so all that logic still runs

src/services/cleanFormSubmissionModel.ts
```diff
@@ -218,19 +216,21 @@ function cleanElementValue(
...
+          case 'textarea':
+          case 'email':
+          case 'barcodeScanner':
+          case 'telephone': {
+            const value = submission[element.name] as string
+            model[element.name] = value.trim()
+            break
+          }
+        }
       }
```
Zaxist, line 234: This is going to add these values to the model even if they should be hidden

src/services/cleanFormSubmissionModel.ts
```diff
@@ -219,16 +219,16 @@ function cleanElementValue(
...
-            const value = submission[element.name] as string
-            model[element.name] = value.trim()
-            break
+          switch (element.type) {
+            case 'text':
+            case 'textarea':
+            case 'email':
+            case 'barcodeScanner':
+            case 'telephone': {
+              const value = submission[element.name] as string
```
Zaxist, line 228: Why not declare this line before the switch, and then use a default case to catch all the other elements to do

```
model[element.name] = submission[element.name]
```
Instead of assigning it twice?

PR #260, opened 2022-03-22
ON-19952 # fixed elements under a section in repeatable sets not bein…

aaronroworth:
…g pre-filled

### DIFF
--- a/src/services/generate-default-data.ts
+++ b/src/services/generate-default-data.ts
@@ -200,7 +200,11 @@ function parseFormSubmissionModel(
   return parseUnknownAsRecord(value, (record) => {
     return elements.reduce<FormSubmissionModel>((model, element) => {
       switch (element.type) {
-        case 'section':
+        case 'section': {
+          const partialModel = parseFormSubmissionModel(element.elements, model)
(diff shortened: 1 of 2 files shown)


divporter:
LGTM

src/services/generate-default-data.ts
```diff
@@ -200,7 +200,11 @@ function parseFormSubmissionModel(
   return parseUnknownAsRecord(value, (record) => {
     return elements.reduce<FormSubmissionModel>((model, element) => {
       switch (element.type) {
-        case 'section':
+        case 'section': {
+          const partialModel = parseFormSubmissionModel(element.elements, model)
+          Object.assign(model, partialModel)
+          break
+        }
```
mymattcarroll, line 207:
```suggestion
        case 'section':
```

aaronroworth: that would just change it back to what it was originally?

mymattcarroll: ah, lol. 

If sections are broken, wouldn't the same bug be there for pages as well?

Can you please confirm the same bug does not exist for pages?

PR #269, opened 2022-04-26
ON-20795 # Added form store components

Zaxist:


src/hooks/useSubmissionIdIsValid.ts
```diff
@@ -0,0 +1,14 @@
+import * as React from 'react'
+import { ValidatorForm } from 'react-material-ui-form-validator'
+
+const submissionIdInvalidMessage = 'Must be a valid Submission Id'
+
+const useSubmissionIdValidationMessage = (submissionId: string | undefined) => {
+  return React.useMemo(() => {
+    if (!ValidatorForm.getValidator('isUUID', submissionId, true)) {
```
mymattcarroll, line 8: As this is a custom validator that we wrote, can we not include this library here and just do the regex check here.

mymattcarroll:
Does the new component need to be added to `index.ts` as well.

src/hooks/useSubmissionIdIsValid.ts
Zaxist: Ah, thought it was built in

PR #289, opened 2022-06-01
ON-21918 # Cleared sorting when turning off unwinding

Zaxist:


src/components/formStore/OneBlinkFormStoreColumnsButton.tsx
```diff
@@ -69,6 +69,9 @@ function OneBlinkFormStoreColumnsButton(
                     onChangeParameters(
                       (currentParameters) => ({
                         ...currentParameters,
+                        sorting: checked
+                          ? currentParameters.sorting
+                          : undefined,
```
mymattcarroll, line 74: Should this be set back to the date time submitted?

Zaxist: Fixed in the next one

PR #317, opened 2022-09-26
ON-25458 # Added loading bar to attachments

Zaxist:


kizaonline:
Just need a changelog entry

PR #326, opened 2022-10-05
ON-25464 # Add prompt to continue with attachments uploading

kizaonline:


src/OneBlinkFormBase.tsx
```diff
@@ -914,6 +924,45 @@ function OneBlinkFormBase({
...
+                <>
+                  <span style={{ flex: 1 }}></span>
+                  <button
+                    type="button"
+                    className="button ob-button is-light cypress-attachments-confirm-wait"
+                    onClick={handleWaitForAttachments}
+                  >
+                    <CustomisableButtonInner
+                      label={'Wait'}
+                      icon={buttons?.cancelPromptNo?.icon}
```
mymattcarroll, line 943: If we are going to add customisable icons here, we prob want to add them to the console, not sure these icons will suit this purpose as well.

PR #327, opened 2022-10-05
ON-25968 # Fixed Data Manager crashing if submission data does not ma…

mymattcarroll:
…tch form element

### DIFF
--- a/src/components/formStore/table/FormElementTableCell.tsx
+++ b/src/components/formStore/table/FormElementTableCell.tsx
@@ -406,7 +406,7 @@ function FormElementTableCell({ formElement, submission, allowCopy }: Props) {
     case 'image':
     case 'heading':
     case 'summary': {
-      break
+      return null
     }
 
     default: {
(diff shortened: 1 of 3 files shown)


src/components/formStore/table/FormElementTableCell.tsx
```diff
@@ -172,25 +178,26 @@ function FormElementTableCell({ formElement, submission, allowCopy }: Props) {
         formElement.type === 'checkboxes' ||
         (formElement.type === 'select' && formElement.multi)
       ) {
-        const value = unknown as string
-        if (!Array.isArray(value) || !value.length) {
+        if (!Array.isArray(unknown) || !unknown.length) {
```
kizaonline, line 181: Would repeatableSet data still pass this check?

mymattcarroll: checking now

kizaonline (approved):

PR #328, opened 2022-10-06
ON-26502 # Pass backgroundUpload to onSaveDraft()

kizaonline:


src/OneBlinkFormBase.tsx
```diff
@@ -53,6 +53,7 @@ export type BaseProps = {
   attachmentRetentionInDays?: number
   onSaveDraft?: (
     newDraftSubmission: submissionService.NewDraftSubmission,
+    backgroundUpload?: boolean,
```
mymattcarroll, line 56: I know this is a bit of a hassle but can we please add this to the `NewDraftSubmission` in apps.

PR #329, opened 2022-10-06
ON-26041 # Dont prompt users authenticated via keys

kizaonline:


src/OneBlinkFormBase.tsx
```diff
@@ -147,6 +147,10 @@ function OneBlinkFormBase({
   // #region Unsaved Changed
 
   const history = useHistory()
+  const isKeyUser = React.useMemo(() => {
+    const token = authService.getUserToken()
```
mymattcarroll, line 151: This should be using the hook `useAuth` which returns a property `isUsingFormsKey`

PR #345, opened 2023-01-19
ON-28846 # Added `isInfoPage` prop

Zaxist:


src/services/determineIsInfoPage.ts
```diff
@@ -0,0 +1,15 @@
...
+
+const infoPageElements: FormTypes.FormElementType[] = [
+  'heading',
+  'html',
+  'image',
+  'section',
+]
+export default function determineIsInfoPage(form: FormTypes.Form) {
+  formElementsService.forEachFormElement(form.elements, (e) => {
+    if (!infoPageElements.includes(e.type)) return false
```
mymattcarroll, line 12: I dont think this will work. this return statement is in an function that is not the `determineIsInfoPage` function.

Using the `findFormElement()` element would be more preformant as it will stop when it finds an element that matches the predicate, instead of iterating over all of the elements.

Zaxist: Pretty sure it will work, but your right about the second part.

PR #360, opened 2023-02-23
ON-29630 # Added mustache rendering to info elements

Zaxist:


src/form-elements/FormElementHTML.tsx
```diff
@@ -2,24 +2,40 @@ import * as React from 'react'
...
+    html = html.replace('{INDEX}', (index + 1).toString())
+
+    html = submissionService.replaceElementValues(html, {
+      submission: formSubmissionModel,
+      formElements: elements,
+      formatCurrency: localisationService.formatCurrency,
+      formatDate: (v) => localisationService.formatDate(new Date(v)),
+      formatNumber: localisationService.formatNumber,
+      formatTime: (v) => localisationService.formatTime(new Date(v)),
+    })
```
divporter, line 27: I'm pretty sure this won't handle conditionally hidden elements, I don't know how fast the function is to clean the submission model, but if it's not that performant in monkey makers land we use the conditionally shown elements to determine if an element is hidden/visible.

Zaxist: Are you saying it will be slow, or it won't work?

divporter: I'm saying the clean the submission model might be slow, but I don't know what it does under the hood.

divporter: And I'm saying that this implementation won't handle conditionally hidden elements

Zaxist: I think it does, I'll double check on it though

Zaxist: Confirmed working.

divporter: 🎉

PR #365, opened 2023-03-08
ON-29929 # Fixed margin on buttons from CSS

jdawg093:


src/styles/buttons.scss
```diff
@@ -0,0 +1,7 @@
+.buttons {
+  gap: 0 0.5rem;
+}
+
+.buttons .button:not(:last-child):not(.is-fullwidth) {
+  margin-right: 0;
+}
```
divporter, line 7: can this be deleted? or will it have margin if not set to zero?

divporter: I am idiot

PR #383, opened 2023-03-20
ON-30278 # Using REM instead of pixels for hint text

jdawg093:


src/styles/renderer.scss
```diff
@@ -80,7 +80,7 @@
 }
 
 .ob-hint-text {
-  font-size: 14px;
+  font-size: 0.875rem;
```
divporter, line 83: could you try with $size-7 and see how it looks?

PR #392, opened 2023-03-27
ON-30467 # add validation css class to repeatable set element

divporter:


src/form-elements/FormElementRepeatableSet.tsx
```diff
@@ -109,8 +109,16 @@ function FormElementRepeatableSet({
     formElementConditionallyShown.type === 'repeatableSet'
       ? formElementConditionallyShown.entries
       : {}
+
+  let validationClass = ''
+  if (displayValidationMessage) {
+    validationClass = formElementValidation
+      ? 'ob-repeatable-set-element__invalid'
+      : 'ob-repeatable-set-element__valid'
+  }
```
Zaxist, line 118: Could do 
```
const validationClass = displayValidationMessage && formElementValidation ? 'ob-repeatable-set-element__invalid' : 'ob-repeatable-set-element__valid'
```

Up to you though

mymattcarroll: Dave, your logic will result in no class unless `displayValidationMessage` is `true`.

Zac's logic will result in a class always being added.

The section element works like Zac's proposal so I would suggested implementing it (if the section element logic is what we are going for)

divporter: I don't like the idea of showing the valid class when the user has not done anything.  Which is the way section elements work, although the implementation may be a bit different.  When the form opens there should be neither valid nor invalid class on the element.

divporter: ie section elements have no classes when the form loads

mymattcarroll: The section element will only show the `valid` class if everything in the section is valid. If you navigate to the form and the section element contains required elements that have no value, it will not have `valid` class. It will also not have the `invalid` class until `displayValidationMessage` is true (which is trigger but navigating to another page, pressing submit, or collapsing the section). Thinking this through, I don't think either of your proposals matches the section element logic 🤔.

Zaxist: That sounds like Dave's code would do just that.

Zaxist: Or is it not doing this part 

> If you navigate to the form and the section element contains required elements that have no value, it will not have valid class.

divporter: Dave's code will do it

mymattcarroll: Daves code does this:

> The `valid` class will only be applied if `displayValidationMessage` is true (which is triggered by the user navigating to another page or pressing submit)

The logic we want is

> The `valid` class will be applied if everything in the section is valid (which is determined by `formElementValidation` not having a value)

mymattcarroll: I believe the logic we want could be implemented like so:

```
const validationClass = React.useMemo(() => {
  if (!formElementValidation) {
    return 'ob-repeatable-set-element__valid'
  }
  if (displayValidationMessage) {
    return 'ob-repeatable-set-element__invalid'
  }
}, [])
```

divporter: Ah yeah, better update the others then.

divporter: Sounds like a job for a hooky hook.

src/components/renderer/OneBlinkFormElements.tsx
```diff
@@ -136,10 +136,13 @@ function OneBlinkFormElements<T extends FormTypes._NestedElementsElement>({
...
         let validationClass = ''
-        if (displayValidationMessages && !('elements' in element)) {
-          validationClass = formElementsValidation?.[element.name]
-            ? 'ob-element__invalid'
-            : 'ob-element__valid'
+        if (!('elements' in element)) {
+          if (!formElementsValidation?.[element.name]) {
+            validationClass = 'ob-element__valid'
+          }
+          if (displayValidationMessages) {
```
Zaxist, line 143: Won't this overwrite the above? They are not mutually exclusive are they?

src/hooks/useValidationClass.ts
```diff
@@ -0,0 +1,31 @@
...
+  formElementsValid: boolean
+  displayValidationMessage: boolean
+  validClassName: string
+  invalidClassName: string
+}): {
+  validationClassName: string
+  valid: boolean
+} {
+  return React.useMemo(() => {
+    if (!formElementsValid) {
```
Zaxist, line 18:
```suggestion
    if (formElementsValid) {
```
Should be the other way around, no?

src/form-elements/FormElementSection.tsx
```diff
@@ -33,16 +34,15 @@ function FormElementSection<T extends FormTypes._NestedElementsElement>({
...
-  const isInvalid = React.useMemo(() => {
-    return (
-      displayValidationMessage &&
-      checkSectionValidity(element, props.formElementsValidation)
-    )
-  }, [displayValidationMessage, element, props.formElementsValidation])
-
-  const isValid = React.useMemo(() => {
-    return !checkSectionValidity(element, props.formElementsValidation)
-  }, [element, props.formElementsValidation])
```
mymattcarroll, line 45: I would leave this hook here and pass `isValid` as `formElementsValid` as the `checkSectionValidity()` could contain multiple iterations, it will benefit from being memo'd.

src/components/renderer/OneBlinkFormElements.tsx
divporter: yeah I fixed this badness

src/hooks/useValidationClass.ts
divporter: Also fixed that badness

src/form-elements/FormElementSection.tsx
divporter: Good idea

PR #393, opened 2023-03-29
ON-30538 # Fixed validation not displaying on unvisited pages when su…

Zaxist:
…bmitting


src/form-elements/FormElementText.tsx
```diff
@@ -29,6 +29,7 @@ function FormElementText({
   const text = typeof value === 'string' ? value : ''
   const isDisplayingValidationMessage =
     (isDirty || displayValidationMessage) && !!validationMessage
+
```
Zaxist, line 32:
```suggestion
```

PR #396, opened 2023-03-30
ON-30555 # move isDirty prop up to FormElementSwitch

divporter:


src/components/renderer/OneBlinkFormElements.tsx
```diff
@@ -195,18 +223,10 @@ const FormElementSwitch = React.memo(function OneBlinkFormElement({
...
-  formElementValidation: FormElementValidation | undefined
-  displayValidationMessage: boolean
-  formElementConditionallyShown: FormElementConditionallyShown | undefined
-  id: string
-  isEven: boolean | undefined
-  onChange: FormElementValueChangeHandler
-  onLookup: FormElementLookupHandler
-}) {
+  isDirty,
+  setIsDirty,
```
Zaxist, line 227: Can just do

```suggestion
 ...dirtyProps,
```
to avoid recreating an object every render.

Zaxist: Don't commit that btw, would need to manually change it

src/form-elements/FormElementCheckBoxes.tsx
```diff
@@ -88,7 +88,9 @@ function FormElementCheckboxes({
               options={filteredOptions}
               selected={selectedValues}
               disabled={element.readOnly}
-              onChange={onChange}
+              onChange={(element, value) => {
+                onChange(element, value)
```
Zaxist, line 92: ?

src/form-elements/FormElementCheckBoxes.tsx
```diff
@@ -102,7 +104,9 @@ function FormElementCheckboxes({
                       element={element}
                       option={option}
                       isSelected={isSelected}
-                      onClick={() => changeValues(option.value, isSelected)}
+                      onClick={() => {
+                        changeValues(option.value, isSelected)
```
Zaxist, line 108: Syntax preference change?

src/form-elements/FormElementCompliance.tsx
```diff
@@ -243,6 +247,8 @@ function FormElementCompliance({
                 validationMessage={undefined}
                 value={typedValue?.notes}
                 element={notesElement}
+                isDirty={isTextDirty}
```
Zaxist, line 250: I think this means that potentially you could have a situation where The invalid class is added to the whole element because the options are invalid, but the notes and files element do not yet show their appropriate validation... but I don't think its too big of a deal.

src/hooks/attachments/useAttachments.ts
```diff
@@ -6,16 +6,15 @@ import {
...
 import { attachmentsService } from '@oneblink/apps'
 import { canvasToBlob } from '../../services/blob-utils'
-import useBooleanState from '../useBooleanState'
 import useIsMounted from '../useIsMounted'
 import { FormElementValueChangeHandler } from '../../types/form'
 
 const useAttachments = (
   element: FormTypes.FilesElement,
   onChange: FormElementValueChangeHandler<attachmentsService.Attachment[]>,
+  setIsDirty: () => void,
```
Zaxist, line 15: Prob best to import this from the type

src/components/renderer/OneBlinkFormElements.tsx
```diff
@@ -184,6 +174,44 @@ function OneBlinkFormElements<T extends FormTypes._NestedElementsElement>({
...
+        return 'ob-element__valid'
+      }
+      if (isDirty || displayValidationMessage) {
+        return 'ob-element__invalid'
+      }
+    }
+  }
+  const validationClassName = getValidationClass()
+
+  if (element.type === 'page' || element.type === 'section') {
```
mymattcarroll, line 195: this seems wrong, should this only be returning null for `'page'` element?

src/form-elements/FormElementCheckBoxes.tsx
divporter: ah yeah the next line got removed so I'll set that back

src/form-elements/FormElementCompliance.tsx
divporter: Yeah, it's really there just to satisfy the props, I didn't want them optional/undefined

src/hooks/attachments/useAttachments.ts
divporter: yeah agree

src/components/renderer/OneBlinkFormElements.tsx
divporter: section and pages have been handled in the component above, this is really just for TS

src/components/renderer/OneBlinkFormElements.tsx
divporter: if other props are added later they'll get mixed up in it, I'll make it a memo

src/components/renderer/OneBlinkFormElements.tsx
Zaxist: Yes, I meant to check that further down, to see if it had just been moved. Didn't appear so, I think this is wrong too.

Zaxist: Sorry, comment posted late.

Zaxist: Could we type the element as excluding those 2 instead?

src/form-elements/FormElementCompliance.tsx
Zaxist: Would it be better to just pass on the original, getter/setters instead of creating new dummy ones?

src/components/renderer/OneBlinkFormElements.tsx
Zaxist: If other props get added later, theres work to be done to pass them on anyway, so doesn't matter IMO, but happy to have a memo.

src/form-elements/FormElementCompliance.tsx
divporter: Yeah, I can see that.  It's not actually possible to make them dirty until the radio is touched anyway, but I guess for future us this would be better.

PR #404, opened 2023-04-04
ON-30665 # Offset Days now working with element Ids Dates

jdawg093:


src/services/generate-default-data.ts
```diff
@@ -104,6 +106,9 @@ export const generateDate = ({
         const offset = date.getTimezoneOffset()
         return new Date(date.getTime() + offset * 60000)
       }
+      if (dateElementIdUsed && daysOffset !== undefined) {
```
mymattcarroll, line 109: As discuss, we dont need this, we can just add the offset to the date if it is passed in.

PR #407, opened 2023-04-06
ON-30671 # Added date to and from validation based on other form elem…

mymattcarroll:
…ents

### DIFF
--- a/src/services/form-validation.ts
+++ b/src/services/form-validation.ts
@@ -1,4 +1,4 @@
-import validate from 'validate.js'
+import validate, { ValidatorConstraintFn } from 'validate.js'
 import { attachmentsService, localisationService } from '@oneblink/apps'
 import { FormTypes } from '@oneblink/types'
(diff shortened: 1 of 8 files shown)


src/services/form-validation.ts
```diff
@@ -291,397 +227,468 @@ const escapeElementName = (elementName: string) => {
...
 
-export function generateValidationSchema(
+function getCleanDateRangeConfiguration(
+  options: DateRangeConfigurationOptions,
   elements: FormTypes.FormElement[],
+  submission: FormSubmissionModel | undefined,
+  formElementsConditionallyShown: FormElementsConditionallyShown | undefined,
+): ReturnType<typeof getDateRangeConfiguration> {
+  if (options.referenceFormElementId && submission) {
+    const { model } = cleanFormSubmissionModel(
```
divporter, line 237: is this potentially a huge performance hit?

PR #417, opened 2023-04-24
ON-31080 # Added logic for reference to other elements for repeatable…

jdawg093:
… sets

### DIFF
--- a/src/services/getRepeatableSetEntriesConfiguration.ts
+++ b/src/services/getRepeatableSetEntriesConfiguration.ts
@@ -1,0 +1,25 @@
+import { FormTypes } from '@oneblink/types'
+import { RepeatableSetElement } from '@oneblink/types/typescript/forms'
+import { FormSubmissionModel } from '../types/form'
+import { submissionService } from '@oneblink/sdk-core'
+
+export default function getRepeatableSetEntriesConfiguration(
+  setEntries: RepeatableSetElement['minSetEntries'],
+  elements: FormTypes.FormElement[],
+  formSubmissionModel: FormSubmissionModel,
+) {
+  if (typeof setEntries === 'object') {
+    const value = submissionService.getRootElementValueById(
+      setEntries.elementId,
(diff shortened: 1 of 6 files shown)


src/services/getRepeatableSetEntriesConfiguration.ts
```diff
@@ -0,0 +1,25 @@
...
+import { RepeatableSetElement } from '@oneblink/types/typescript/forms'
+import { FormSubmissionModel } from '../types/form'
+import { submissionService } from '@oneblink/sdk-core'
+
+export default function getRepeatableSetEntriesConfiguration(
+  setEntries: RepeatableSetElement['minSetEntries'],
+  elements: FormTypes.FormElement[],
+  formSubmissionModel: FormSubmissionModel,
+) {
+  if (typeof setEntries === 'object') {
```
Zaxist, line 11: When you check for `object` type, best to also check it is `!== null` as `null` is classified as `object` type. 
https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Operators/typeof

Zaxist: Obviously this shouldn't be null anyway, but its still a good idea when writing these kinds of type guards.

src/services/getRepeatableSetEntriesConfiguration.ts
```diff
@@ -0,0 +1,25 @@
+import { FormTypes } from '@oneblink/types'
+import { RepeatableSetElement } from '@oneblink/types/typescript/forms'
+import { FormSubmissionModel } from '../types/form'
+import { submissionService } from '@oneblink/sdk-core'
+
+export default function getRepeatableSetEntriesConfiguration(
```
Zaxist, line 6: My only other suggestion here would be to just make this function a hook, and return the `useMemo` from it. Rather than having to wrap it twice inside a `useMemo` in the component.

PR #418, opened 2023-04-24
Draft: ON-31087 # Fixed isDirty validation for Checkbox element

Zaxist:


src/form-elements/FormElementCheckBoxes.tsx
```diff
@@ -34,12 +35,39 @@ function FormElementCheckboxes({
...
+      (element.required && !!selectedValues.length) ||
+      // or all options are selected and requiredAll is true
+      (element.requiredAll && selectedValues.length === element.options?.length)
+    ) {
+      setIsDirty()
+    }
+    // eslint-disable-next-line react-hooks/exhaustive-deps
+  }, [
+    // `isDirty` is excluded from this dep list because we do not want to trigger re-runs
+    // when it changes, it is only referenced to short circuit the function when it does run
```
mymattcarroll, line 63: You should be able to make `isDirty` a dep here, it will only change once from `false` to `true` and when it does it will not do anything as the comment suggests.

Zaxist: I know it can be included, I just don't want to evaluate this again, when there is no reason for it to be evaluated again.

mymattcarroll: I get it :)

Biggest downside is that `eslint-disable-next-line react-hooks/exhaustive-deps` applies to all deps, not just that one. So if another dep was added in the future as part of a refactor, the implementor might not add the new dep because eslint did not warn them to add it.

If you want to keep it out, im happy 👍

PR #421, opened 2023-05-01
ON-31212 # Fixed default value for date not catering for timezone

mymattcarroll:


divporter:
Approved: https://github.com/oneblink/apps-react/pull/422/files
This PR no longer necessary

PR #436, opened 2023-05-26
ON-31943 # `useSignUp` hook

RyanButton:


kizaonline:
`useFormsAppUserSignUp.ts`?

RyanButton:
I don't mind that name, I just wanted it to mirror `useLogin.ts` is all

src/hooks/useSignUp.ts
```diff
@@ -0,0 +1,91 @@
...
+      signUpError: null,
+    }))
+
+    try {
+      const newUser = await authService.signUp({
+        formsAppId,
+        email: username,
+        firstName,
+        lastName,
+        generatePassword: true,
```
kizaonline, line 66: Whys this always true?

RyanButton: a doozie

kizaonline (approved):

PR #450, opened 2023-06-08
ON-32308 # Added usePendingSubmissions hook

mymattcarroll:


src/hooks/usePendingSubmissions.tsx
```diff
@@ -0,0 +1,263 @@
...
+  React.useEffect(() => {
+    if (submittedNotificationCount > 0) {
+      const timeoutId = setTimeout(() => {
+        setSubmittedNotificationCount(0)
+      }, 5000)
+      return () => {
+        clearTimeout(timeoutId)
+      }
+    }
+  }, [submittedNotificationCount])
```
divporter, line 218: Could this be done better, ie. without the optimistic timer? (disclaimer: I don't know all the ins and outs)

mymattcarroll: I think ill make the 5 seconds an input prop but default it to 5000. But I would prefer to keep this here as it would be quite annoying to implement with the `submittedNotificationCount` variable being a number.

PR #455, opened 2023-06-09
ON-32346 # store additional parameters in localStorage

divporter:


src/components/formStore/FormStoreTableProvider.tsx
```diff
@@ -48,6 +57,11 @@ export function FormStoreTableProvider({
     debounceSearchMs: 1000,
     onDefaultFilters: React.useCallback((query) => {
       let defaultParameters: formStoreService.FormStoreParameters = {}
+      const localStorageParams = getParamsFromLocalStorage()
+      if (localStorageParams?.unwindRepeatableSets === true) {
```
mymattcarroll, line 61:
```suggestion
      if (localStorageParams?.unwindRepeatableSets) {
```

src/components/formStore/FormStoreTableProvider.tsx
```diff
@@ -109,6 +123,16 @@ export function FormStoreTableProvider({
...
 
+  React.useEffect(() => {
+    const paramsToStore = {
+      unwindRepeatableSets: parameters?.unwindRepeatableSets,
+    }
+    localStorage.setItem(
+      TABLE_PARAMETERS_FORM_STORE_KEY,
+      JSON.stringify(paramsToStore),
+    )
+  }, [parameters])
```
mymattcarroll, line 134:
```suggestion
  }, [parameters?.unwindRepeatableSets])
```

src/components/formStore/FormStoreTableProvider.tsx
Zaxist: `defaultParameters` is getting overwritten below.

divporter: I didn't want to do a truthy match in case the value contained something like "dog" - not that it can unless the user goes in and changes it.  But I felt safer this way

divporter: @Zaxist true, will fix

mymattcarroll (approved):

PR #464, opened 2023-06-19
ON-32622 # Changed lookups to stop when input value changes

mymattcarroll:


src/hooks/useLookupNotification.ts
```diff
@@ -24,22 +28,37 @@ export default function useLookupNotification(value: unknown) {
...
+    onLookup({
+      newValue: value,
+      abortController: abortControllerRef.current,
+      continueLookupOnAbort: false,
+    })
     // Wants to use "onLookup" and "value" as dependencies,
     // however, these will change on any change made on any
     // element. Checking if "lookupCount" has changed is enough
     // to trigger a lookup when the correct dependencies change
     // eslint-disable-next-line react-hooks/exhaustive-deps
```
divporter, line 50: now that I'm a react docs aficionado,  it would be cool to write this without the disable, but equally this has worked for a long time. Up to you (not a challenge)

divporter (approved):

src/hooks/useLookupNotification.ts
mymattcarroll: I am ashamed to say this has already cost me a few hours sleep and I can't work out a way to do this without a bunch of refs (I prefer this).

PR #466, opened 2023-06-22
ON-32726 # Changed Payment Receipt Items to be returned from apps

jdawg093:


src/PaymentReceipt.tsx
```diff
@@ -17,6 +16,7 @@ import {
   ReceiptListItem,
   ReceiptButton,
 } from './components/receipts'
+import { PaymentReceiptItem } from '@oneblink/apps/dist/types/payments'
```
mymattcarroll, line 19: This should be export as on paymentService as `paymentService.PaymentReceiptItem`. Ill fix apps now.

mymattcarroll (approved):

PR #475, opened 2023-07-12
ON-33060 # Clears label when lookup prefills undefined on autocomplete

jdawg093:


src/components/renderer/AutocompleteDropdown.tsx
```diff
@@ -99,7 +99,7 @@ function AutocompleteDropdown<T>({
     setError(null)
     onClose()
 
-    if (!value && Array.isArray(options)) {
```
Zaxist, line 102: Shouldn't this still only happen when there is no value?

Zaxist: nvm

src/components/renderer/AutocompleteDropdown.tsx
```diff
@@ -114,15 +114,16 @@ function AutocompleteDropdown<T>({
         }
```
Zaxist, line 114: I think we probably only want the `onSelectOption` to happen if that option is not already selected.

src/components/renderer/AutocompleteDropdown.tsx
```diff
@@ -71,6 +71,58 @@ function AutocompleteDropdown<T>({
...
+    },
+    [
+      label,
+      onChangeLabel,
+      onChangeValue,
+      onClose,
+      onSelectOption,
+      options,
+      setIsDirty,
+      value,
```
mymattcarroll, line 122: Does this still need `value` and `label` as dependencies?

jdawg093: It does not. Updating with the change now :)

src/components/renderer/AutocompleteDropdown.tsx
```diff
@@ -71,6 +71,52 @@ function AutocompleteDropdown<T>({
...
+            if (!value) {
+              console.log('Setting value after blurring away from autocomplete')
+              onSelectOption(option)
+            }
+            return
+          }
+        }
+
+        console.log('Removing label after blurring away from autocomplete')
+        onChangeValue(undefined)
```
mymattcarroll, line 105: This should not be required.

src/components/renderer/AutocompleteDropdown.tsx
```diff
@@ -71,6 +71,52 @@ function AutocompleteDropdown<T>({
...
+      if (Array.isArray(options)) {
+        // If there is no option currently selected but the typed in label
+        // matches an option's label, set that option as the value, otherwise remove label
+        if (inputOption.label) {
+          const lowerCase = inputOption.label.toLowerCase()
+          const option = options.find(
+            (option) => option.label.toLowerCase() === lowerCase,
+          )
+          if (option) {
+            if (!value) {
```
mymattcarroll, line 96: Why is this here?

src/components/renderer/AutocompleteDropdown.tsx
mymattcarroll: ignore

src/components/renderer/AutocompleteDropdown.tsx
mymattcarroll: ignore

src/components/renderer/AutocompleteDropdown.tsx
```diff
@@ -71,6 +71,51 @@ function AutocompleteDropdown<T>({
...
+      if (Array.isArray(options)) {
+        // If there is no option currently selected but the typed in label
+        // matches an option's label, set that option as the value, otherwise remove label
+        if (inputOption.label) {
+          const lowerCase = inputOption.label.toLowerCase()
+          const option = options.find(
+            (option) => option.label.toLowerCase() === lowerCase,
+          )
+          if (option) {
+            if (!inputOption.value) {
```
mymattcarroll, line 96:
```suggestion
            if (inputOption.value !== option.value) {
```

src/form-elements/FormElementAutocomplete.tsx
```diff
@@ -103,6 +103,7 @@ const AutocompleteFilter = React.memo(function AutocompleteFilter({
     if (option && label !== option.label) {
       setLabel(option.label)
     }
+
```
mymattcarroll, line 106:
```suggestion
```

PR #476, opened 2023-07-13
ON-33045 # Changed form element to display options as soon as possible

mymattcarroll:


src/hooks/useDynamicOptionsLoaderState.tsx
```diff
@@ -0,0 +1,333 @@
...
+      <FormElementOptionsContext.Provider value={optionsSetResults}>
+        <FreshdeskFieldsStateContext.Provider value={freshdeskFieldsState}>
+          {children}
+        </FreshdeskFieldsStateContext.Provider>
+      </FormElementOptionsContext.Provider>
+    </>
+  )
+}
+
+const LoadOptionsSet = React.memo(function LoadOptionsSet({
```
Zaxist, line 145: What's the point of this one being a component?

mymattcarroll: so the effect can have the individual OptionSetResult as a dep instead of the entire array

PR #483, opened 2023-07-28
ON-33486 # Added support for static lookups to work offline

jdawg093:


src/components/renderer/LookupNotification.tsx
```diff
@@ -141,6 +141,14 @@ function LookupNotificationComponent({
     [element, injectPagesAfter, onLookup],
   )
 
+  const isNotStaticLookup = React.useMemo(() => {
+    return (
+      (formElementDataLookup && formElementDataLookup.type !== 'STATIC_DATA') ||
+      (formElementElementLookup &&
+        formElementElementLookup.type !== 'STATIC_DATA')
```
divporter, line 148: odd that TS thinks this could be anything else but an element lookup

jdawg093: Bad logic setup on my behalf
We use the same type for both of them though and the type property indicates what it is, I wrote it more as a safeguard (and maybe future proofing if we are to add Element Static Lookups :o )

jdawg093: Could change it to only check for if the elementLookup exists as opposed to checking the type too?

PR #484, opened 2023-08-01
ON-33546 # remove element from executedLookups if value changes

divporter:


src/components/renderer/LookupNotification.tsx
```diff
@@ -86,12 +84,19 @@ function LookupNotificationComponent({
...
+      elementLookupResult,
+      executedLookup,
+    }: {
+      newValue: unknown
+      dataLookupResult: SubmissionTypes.S3SubmissionData['submission']
+      elementLookupResult: FormTypes.FormElement[]
+      executedLookup: ExecutedLookups
+    }) => {
+      const executedLookupResult = executedLookup?.[element.name]
+      if (elementLookupResult && executedLookupResult !== false) {
```
mymattcarroll, line 99: Whats the intention here? We are only injecting pages if the lookup ran successfully in the past?

divporter: That is the arg it receives, so it's the current value.  So only injecting if the lookup was successful

src/form-elements/FormElementRepeatableSet.tsx
```diff
@@ -243,9 +306,16 @@ const RepeatableSetEntry = React.memo<RepeatableSetEntryProps>(
...
       useBooleanState(false)
 
-    const handleChange = React.useCallback(
-      (element: FormTypes.FormElement, value: unknown) => {
-        onChange(index, element, value)
+    const handleChange: NestedFormElementValueChangeHandler = React.useCallback(
+      (nestedElement, { value, executedLookups }) => {
+        if (!('name' in nestedElement)) {
+          return
+        }
```
mymattcarroll, line 313: is this needed?

divporter: Not anymore, nice spot

src/services/form-validation.ts
```diff
@@ -167,7 +180,11 @@ validate.validators.lookups = function (
     return
   }
 
-  if (elementIdsWithLookupsExecuted.includes(formElement.id)) {
+  const elementExecutedLookups = executedLookups?.[formElement.name]
+  if (
+    typeof elementExecutedLookups === 'boolean' &&
```
mymattcarroll, line 185: dont need this

src/services/form-validation.ts
```diff
@@ -623,6 +639,15 @@ export function generateValidationSchema(
             submission,
             formElementsConditionallyShown,
           )
+
+          const repeatableSetExecutedLookups =
+            executedLookups !== undefined &&
+            typeof executedLookups !== 'boolean' &&
+            !Array.isArray(executedLookups) &&
+            Array.isArray(executedLookups[formElement.name])
```
mymattcarroll, line 647: Not sure if this would work but would likely be faster

```suggestion
            formElement.name in  executedLookups
```

src/services/form-validation.ts
```diff
@@ -623,6 +639,15 @@ export function generateValidationSchema(
             submission,
             formElementsConditionallyShown,
           )
+
+          const repeatableSetExecutedLookups =
+            executedLookups !== undefined &&
+            typeof executedLookups !== 'boolean' &&
+            !Array.isArray(executedLookups) &&
```
mymattcarroll, line 646: Not sure if this will work, but will likely be faster

```suggestion
             formElement.name in executedLookups &&
```

divporter: Won't work without casting, which I didn't want to do

PR #486, opened 2023-08-03
ON-33573 # mark element lookup having run as failed if value changed …

divporter:
…by a lookup

### DIFF
--- a/src/components/renderer/LookupNotification.tsx
+++ b/src/components/renderer/LookupNotification.tsx
@@ -83,6 +83,56 @@ function LookupNotificationComponent({
     formElementElementLookup?.runLookupOnClear,
   ])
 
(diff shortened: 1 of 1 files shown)


src/components/renderer/LookupNotification.tsx
```diff
@@ -143,14 +193,18 @@ function LookupNotificationComponent({
             [element.name]: newValue,
             ...dataLookupResult,
           }),
-          executedLookups: {
-            ...executedLookups,
-            ...executedLookup,
-          },
+          executedLookups: mergeExecutedLookups({
+            dataLookupResult,
+            currentSubmission: model,
```
mymattcarroll, line 198:
```suggestion
            currentSubmission: submission,
```

PR #513, opened 2023-10-20
ON-35569 # Added dates to calculations

jdawg093:


src/form-elements/FormElementCalculation.tsx
```diff
@@ -84,6 +84,28 @@ function FormElementCalculation({ element, onChange, value }: Props) {
...
+                  elementValue,
+                  "yyyy-MM-dd'T'HH:mm:ss.SSSX",
+                  new Date(),
+                )
+                const parsedDate = parse(elementValue, 'yyyy-MM-dd', new Date())
+                if (
+                  (isDate(parsedIsoDate) && !isNaN(parsedIsoDate.getDate())) ||
+                  (isDate(parsedDate) && !isNaN(parsedDate.getDate()))
+                ) {
+                  return Date.parse(elementValue)
```
mymattcarroll, line 106: This should ideally be using https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Date/getTime from the parse dates.

src/form-elements/FormElementCalculation.tsx
```diff
@@ -84,6 +84,28 @@ function FormElementCalculation({ element, onChange, value }: Props) {
...
+
+                // Date-fns has a parseIso function, but it'll interpret '10', '11', etc.
+                // as an iso date and cause issues with these string numbers. To combat this problem,
+                // using the parse function with the ISO format will bypass this.
+                const parsedIsoDate = parse(
+                  elementValue,
+                  "yyyy-MM-dd'T'HH:mm:ss.SSSX",
+                  new Date(),
+                )
+                const parsedDate = parse(elementValue, 'yyyy-MM-dd', new Date())
```
mymattcarroll, line 101: Might be more efficient to check the `parseIsoDate` and return early before before creating the `parsedDate`.

src/form-elements/FormElementCalculation.tsx
```diff
@@ -84,6 +84,28 @@ function FormElementCalculation({ element, onChange, value }: Props) {
...
+                // and as a date in the format below to cover all calculation checks
+                // with 'date', 'datetime' and 'time' elements. If the string is not
+                // one of these, then we want to parse it as a float.
+
+                // Date-fns has a parseIso function, but it'll interpret '10', '11', etc.
+                // as an iso date and cause issues with these string numbers. To combat this problem,
+                // using the parse function with the ISO format will bypass this.
+                const parsedIsoDate = parse(
+                  elementValue,
+                  "yyyy-MM-dd'T'HH:mm:ss.SSSX",
```
mymattcarroll, line 98: Prob better to use https://date-fns.org/v2.30.0/docs/parseISO here instead

jdawg093: The comment above is why I couldn't use that function, I had that initially but led to the issue where "10", "11" and other numbers were being parsed as ISO strings and resulting in these results being parsed as a date

src/form-elements/FormElementCalculation.tsx
```diff
@@ -84,6 +84,28 @@ function FormElementCalculation({ element, onChange, value }: Props) {
...
+                  new Date(),
+                )
+                if (isDate(parsedIsoDate) && !isNaN(parsedIsoDate.getDate())) {
```
mymattcarroll, line 101:
```suggestion
                if (!isNaN(parsedIsoDate.getDate())) {
```

Don't need the check on date as `parsedIsoDate` will always be a date according to the types.

src/form-elements/FormElementCalculation.tsx
```diff
@@ -84,6 +84,28 @@ function FormElementCalculation({ element, onChange, value }: Props) {
...
+                  return new Date(elementValue).getTime()
```
mymattcarroll, line 102:
```suggestion
                  return parsedIsoDate.getTime()
```

src/form-elements/FormElementCalculation.tsx
```diff
@@ -84,6 +84,28 @@ function FormElementCalculation({ element, onChange, value }: Props) {
...
+
+                // Date-fns has a parseIso function, but it'll interpret '10', '11', etc.
+                // as an iso date and cause issues with these string numbers. To combat this problem,
+                // using the parse function with the ISO format will bypass this.
+                const parsedIsoDate = parse(
+                  elementValue,
+                  "yyyy-MM-dd'T'HH:mm:ss.SSSX",
+                  new Date(),
+                )
+                if (!isNaN(parsedIsoDate.getDate())) {
```
mymattcarroll, line 101:
```suggestion
                const parsedIsoDate = localisationService.generateDate({
                  value: elementValue,
                  daysOffset: undefined,
                  dateOnly: false,
                })
                if (parsedIsoDate) {
```

This logic is in apps (now updated to include your date-fns magic).

src/form-elements/FormElementCalculation.tsx
```diff
@@ -84,6 +84,28 @@ function FormElementCalculation({ element, onChange, value }: Props) {
...
+                  return parsedIsoDate.getTime()
+                }
+                const parsedDate = parse(elementValue, 'yyyy-MM-dd', new Date())
```
mymattcarroll, line 104: same here.

src/form-elements/FormElementCalculation.tsx
```diff
@@ -8,7 +8,7 @@ import { localisationService } from '@oneblink/apps'
 import { FormElementValueChangeHandler } from '../types/form'
 import { formElementsService } from '@oneblink/sdk-core'
 import QuillHTML from '../components/QuillHTML'
-import { parse } from 'date-fns'
+import { generateDate } from '@oneblink/apps/dist/localisation-service'
```
mymattcarroll, line 11:
```suggestion
import { localisationService } from '@oneblink/apps'
```

src/form-elements/FormElementCalculation.tsx
```diff
@@ -84,6 +83,33 @@ function FormElementCalculation({ element, onChange, value }: Props) {
...
               if (typeof elementValue === 'string') {
+                // The string could be an iso date string, or a string
+                // resembling a date. We need to parse the value as an ISO string
+                // and as a date in the format below to cover all calculation checks
+                // with 'date', 'datetime' and 'time' elements. If the string is not
+                // one of these, then we want to parse it as a float.
+
+                // Date-fns has a parseIso function, but it'll interpret '10', '11', etc.
+                // as an iso date and cause issues with these string numbers. To combat this problem,
+                // using the parse function with the ISO format will bypass this.
```
mymattcarroll, line 94: this comment is no longer relevant, it could be transferred to apps.

src/form-elements/FormElementCalculation.tsx
```diff
@@ -84,6 +83,33 @@ function FormElementCalculation({ element, onChange, value }: Props) {
...
+
+                const parsedIsoDate = localisationService.generateDate({
+                  value: elementValue,
+                  daysOffset: undefined,
+                  dateOnly: false,
+                })
+                if (parsedIsoDate && !isNaN(parsedIsoDate.getDate())) {
```
mymattcarroll, line 101:
```suggestion
                if (parsedIsoDate) {
```

This check is done in apps, it will never return an invalid date.

src/form-elements/FormElementCalculation.tsx
```diff
@@ -84,6 +83,33 @@ function FormElementCalculation({ element, onChange, value }: Props) {
...
+                  return parsedIsoDate.getTime()
+                }
+                const parsedDate = localisationService.generateDate({
+                  value: elementValue,
+                  daysOffset: undefined,
+                  dateOnly: true,
+                })
+                if (parsedDate && !isNaN(parsedDate.getDate())) {
```
mymattcarroll, line 109:
```suggestion
                if (parsedDate) {
```

mymattcarroll (approved):

PR #515, opened 2023-10-26
ON-35784 # Updated sdk and apps

mymattcarroll:


src/OneBlinkForm.tsx
```diff
@@ -565,14 +566,11 @@ const OneBlinkFormUncontrolled = React.memo(function OneBlinkFormUncontrolled({
   initialSubmission,
   resumeAtElement,
   ...props
-}: OneBlinkFormBaseProps & {
```
divporter, line 568: Have the base props been updated? is that what's going on here?

mymattcarroll: yes, yes.

PR #558, opened 2023-12-18
ON-37123 # Number slider now has focus css

jdawg093:


src/styles/number.scss
```diff
@@ -40,3 +40,15 @@
...
 }
+
+// Works for all other major browser except Firefox
+input[type="range"]:focus-visible::-webkit-slider-thumb {
+  outline-offset: $size-9;
+  outline: $input-focus-border-color auto 3px;
+}
+
+// Works for Firefox
+input[type="range"]:focus-visible::-moz-range-thumb {
```
mymattcarroll, line 51:
```suggestion
input[type="range"]:focus-visible::-webkit-slider-thumb,
// Works for Firefox
input[type="range"]:focus-visible::-moz-range-thumb {
```

PR #559, opened 2023-12-18
ON-37142 # Separated CSS class with comment

jdawg093:


src/styles/number.scss
```diff
@@ -41,8 +41,15 @@
   }
 }
 
+// While these are identical CSS classes, they need to be kept
```
divporter, line 44:
```suggestion
// While these are identical CSS attributes/properties, they need to be kept
```

src/styles/number.scss
```diff
@@ -41,8 +41,15 @@
   }
 }
 
+// While these are identical CSS attributes/properties, they need to be kept
```
jdawg093, line 44:
```suggestion
// While these are identical CSS attributes, they need to be kept
```

PR #571, opened 2024-01-22
ON-37440 # Added payment form component

mymattcarroll:


src/components/payments/WestpacQuickStreamPaymentForm.tsx
```diff
@@ -0,0 +1,384 @@
...
+          isLoading: false,
+          loadError: error as Error,
+        })
+      }
+    }
+
+    run()
+
+    return () => {
+      document.body.appendChild(scriptEle)
```
divporter, line 221: shouldn't this be cleaning up, like removing the event listeners etc

mymattcarroll: I believe you are correct!!!

src/components/payments/WestpacQuickStreamPaymentForm.tsx
```diff
@@ -0,0 +1,402 @@
...
+          })
+        }
+      }
+    }
+
+    run()
+
+    return () => {
+      abortController.abort()
+      document.body.appendChild(scriptEle)
```
divporter, line 239: I meant removing the listeners and deleting this line, because you append the script element upstream already. unless I'm missing something?

mymattcarroll: lol 🤦 This line is meant to remove the node from the dom.

PR #577, opened 2024-01-23
ON-37567 # Attempted to prettyify the payment form

jdawg093:


src/styles/westpac-payment.scss
```diff
@@ -0,0 +1,9 @@
+.payment-form-actions {
+    display: flex;
+    gap: 0.5rem;
```
mymattcarroll, line 3: ideally we use a `$size-*` value here and below.

PR #589, opened 2024-02-01
ON-37804 # Changed radio buttons to container query

Zaxist:


src/styles/radio.scss
```diff
@@ -5,11 +5,12 @@
...
+}
 
-  @media only screen and (max-width: 500px) {
-    .ob-button-radio-container {
-      width: 100%;
-
+@container ob-form-container (max-width: 500px) {
+  .ob-button-radio-container {
+    width: 100%;
+    .ob-buttons-radio {
```
mymattcarroll, line 13: this class if above `.ob-button-radio-container`, right?

Zaxist: No.

mymattcarroll: In the existing CSS `.ob-button-radio-container` is inside  `.ob-buttons-radio`.

PR #609, opened 2024-02-26
ON-38417 # Added support for injected options across repeatable sets

mymattcarroll:


src/services/injectableOptions.ts
```diff
@@ -1,290 +1,177 @@
...
 import { v4 as uuidv4 } from 'uuid'
-import { TaskContext } from '../hooks/useTaskContext'
 import { MiscTypes } from '@oneblink/types'
+import { TaskContext } from '../hooks/useTaskContext'
+
+type Option = Pick<FormTypes.ChoiceElementOption, 'value' | 'label'>
+
+type Match = Map<string, Match | undefined>
+
+function groupElementNames(memo: Match, rootElementNames: string[]): Match {
```
divporter, line 12: why does this bother recurse all the way down when processInjectableDynamicOption will just run it again when it goes down the next layer?

tests/services/injectableOptions.test.ts
```diff
@@ -5,9 +5,84 @@ const otherArgs = {
...
+    expect(options).toEqual([
+      {
+        id: '1edcce9e-3635-407b-8adc-ce614fde742c',
+        label: otherArgs.userProfile.email,
+        value: otherArgs.userProfile.username,
+      },
+    ])
+  })
+
+  it.skip('should return no options if any of the replaceable parameters are missing', () => {
```
divporter, line 36: will be unskipped after core-sdk stuff yeah?

src/services/injectableOptions.ts
mymattcarroll: We need to get the full tree to pass the subtree down recursively

mymattcarroll: Or maybe we don't...Let me try some tests with a short circuit.

PR #611, opened 2024-02-26
ON-38490 # Added support for injected options inside repeatable sets

mymattcarroll:


src/services/injectableOptions.ts
```diff
@@ -158,55 +158,130 @@ export default function processInjectableOption({
...
+  submission,
+  ...params
+}: {
+  elements: FormTypes.FormElement[]
+  submission: SubmissionTypes.S3SubmissionData['submission']
+  taskContext: TaskContext
+  userProfile: MiscTypes.UserProfile | undefined
+}): FormTypes.FormElement[] {
+  // We iterate over entries as an array of submission objects because
+  // elements with options within repeatable sets need to include all
```
Zaxist, line 280: children elements of a repeatable set that have injected options need to include all

src/services/injectableOptions.ts
```diff
@@ -158,55 +158,130 @@ export default function processInjectableOption({
...
+  // of the options from each entry within the set. Otherwise we will
```
Zaxist, line 281:
```suggestion
  // children elements of a repeatable set that have injected options need to include all
  // of the injected options from the same elements in sibling entries within the set. Otherwise we will
```

PR #641, opened 2024-03-24
ON-39137 # changed @arcgis/core to use exact version

aaronroworth:


package.json
```diff
@@ -7,7 +7,7 @@
     "url": "https://github.com/oneblink/apps-react/issues"
   },
   "dependencies": {
-    "@arcgis/core": "^4.28.10",
```
kizaonline, line 10: Can we range it to the patch version?

aaronroworth: I believe so, i'll update

package.json
```diff
@@ -7,7 +7,7 @@
...
+    "@arcgis/core": "4.28.10",
```
kizaonline, line 10:
```suggestion
    "@arcgis/core": "~4.28.10",
```

PR #647, opened 2024-03-28
ON-39056 # stop focusing first element on form

RyanButton:


package-lock.json
```diff
@@ -6,7 +6,7 @@
   "packages": {
     "": {
       "name": "@oneblink/apps-react",
-      "version": "5.12.2-beta.3",
```
kizaonline, line 9: Is this right?

PR #651, opened 2024-04-22
ON-39493 # ReverseGeocode element

divporter:


src/components/renderer/ReverseGeocode.tsx
```diff
@@ -0,0 +1,156 @@
...
+        }
+        setReverseGeocodingState((current) => ({
+          ...current,
+          reverseGeocodingErrorMsg: errorMsg,
+        }))
+      }
+      setReverseGeocodingState((current) => ({
+        ...current,
+        isReverseGeocoding: false,
+      }))
```
Zaxist, line 125: Prob want to have 1 state update in the try and the catch so we aren't doing 2 state updates one after the other in the catch scenario.

PR #658, opened 2024-05-06
ON-39513 # added `FLOOR` and `CEILING` commands to calculation elements

divporter:


src/form-elements/FormElementCalculation.tsx
```diff
@@ -207,6 +207,18 @@ function FormElementCalculation({ element, onChange, value }: Props) {
       }
       return null
     })
+    exprParser.registerFunction('FLOOR', (value: number) => {
+      if (!Number.isNaN(value) && Number.isFinite(value)) {
+        return Math.floor(value)
+      }
+      return null
+    })
+    exprParser.registerFunction('CEILING', (value: number) => {
```
Zaxist, line 216: Should be `ROUND_UP`

src/form-elements/FormElementCalculation.tsx
```diff
@@ -207,6 +207,18 @@ function FormElementCalculation({ element, onChange, value }: Props) {
       }
       return null
     })
+    exprParser.registerFunction('FLOOR', (value: number) => {
```
Zaxist, line 210: Should be `ROUND_DOWN`

PR #666, opened 2024-05-21
ON-40405 # FormElementGoogleAddress

divporter:


src/form-elements/FormElementLocation.tsx
```diff
@@ -372,6 +372,7 @@ const LocationPicker = React.memo(function LocationPicker({
 
   const { isLoaded } = useJsApiLoader({
     googleMapsApiKey: googleMapsApiKey ?? '',
+    libraries: ['maps', 'marker', 'places'],
```
mymattcarroll, line 375: Should we move this hook higher into the dom and make it a provider so it is only happening once?

src/form-elements/FormElementGoogleAddress.tsx
```diff
@@ -0,0 +1,209 @@
...
+              input,
+            },
+            (predictions, status) => {
+              if (
+                status !== google.maps.places.PlacesServiceStatus.OK ||
+                !predictions
+              ) {
+                reject(
+                  google.maps.places.PlacesServiceStatus.OK
+                    ? 'Google Places did not return any suggestions'
```
mymattcarroll, line 71: Shouldn't we be returning an empty array if there are not predictions?

divporter: ah yeah, maka sense

src/form-elements/FormElementLocation.tsx
divporter: just load it in regardless? or check if there is a location/google element in the form?

mymattcarroll: 🤔 will it fail if you attempt to load a lib that the key does not support?

divporter: No it won't on load, tested already.  It will only fail when you try and use it to do something it can't.

mymattcarroll: mmm I think we iterate over the form elements to check if we need to load the script

divporter: we'd have to do that each render in case of element lookups.

divporter: Dun worry, Uncle Dave has an idea

src/form-elements/FormElementGoogleAddress.tsx
```diff
@@ -0,0 +1,196 @@
...
+
+  // Ensure the label is set if the value is set outside of this component
+  React.useEffect(() => {
+    if (value) {
+      const newLabel = value.formattedAddress || value.id
+      if (label !== newLabel) {
+        setLabel(newLabel || '')
+      }
+    }
+    // eslint-disable-next-line react-hooks/exhaustive-deps
```
Zaxist, line 152: Prob want to whack a comment specifically before/on this line to say what we need it.

PR #671, opened 2024-05-23
ON-40464 # add powered by google to autocomplete

divporter:


src/components/renderer/AutocompleteDropdown.tsx
```diff
@@ -259,6 +261,7 @@ function AutocompleteDropdown<T>({
 
   return (
     <>
+      {branding}
```
mymattcarroll, line 264: should this be here?

divporter: Yeah!?

divporter: No

PR #673, opened 2024-05-24
ON-40638 # capture error where permissions are not correct

divporter:


src/form-elements/FormElementGoogleAddress.tsx
```diff
@@ -62,27 +62,35 @@ function FormElementGoogleAddress({
...
-                reject('Google Places service not available')
-                return
+              switch (status) {
+                case google.maps.places.PlacesServiceStatus.ZERO_RESULTS: {
+                  resolve([])
+                  break
+                }
+                case google.maps.places.PlacesServiceStatus.REQUEST_DENIED: {
+                  reject(
+                    'Google Maps API key has not been configured correctly',
```
mymattcarroll, line 72: bad practice to reject/throw anything other than an `Error`.

divporter: yah fair enough, fixing

src/form-elements/FormElementGoogleAddress.tsx
```diff
@@ -62,27 +62,39 @@ function FormElementGoogleAddress({
...
           )
         }).catch((e) => {
           if (!abortSignal.aborted) {
             Sentry.captureException(e)
           }
           throw new OneBlinkAppsError(
-            'An unknown error has occurred. Please contact support if the problem persists.',
-            { originalError: e },
+            e instanceof Error ? e.message : e.toString(),
           )
```
mymattcarroll, line 98: Should just be 

```
`throw e`
```

src/form-elements/FormElementGoogleAddress.tsx
```diff
@@ -162,7 +176,9 @@ function FormElementGoogleAddress({
         if (isMounted.current) {
           setError(
             new OneBlinkAppsError(
```
mymattcarroll, line 178: same here,

PR #682, opened 2024-06-03
ON-40940 # stop auto save from occurring while form is disabled

divporter:


mymattcarroll (changes requested):
LGTM 👍 

Just needs changelog entry

mymattcarroll (approved):

PR #688, opened 2024-06-07
ON-41012 # prevent auto save of disabled form inside throttledAutoSave

divporter:


src/hooks/useFormSubmissionAutoSaveState.ts
```diff
@@ -64,27 +64,36 @@ export default function useFormSubmissionAutoSaveState({
...
-              return autoSaveService.upsertAutoSaveData<FormElement>(
-                definition.id,
-                `LAST_ELEMENT_UPDATED_${autoSaveKey}`,
-                lastElementUpdated,
-              )
+        if (!formIsDisabled) {
+          switch (lastElementUpdated?.type) {
+            case 'summary':
+            case 'calculation':
+            case 'captcha': {
```
mymattcarroll, line 71: This wont prevent the `resetRecaptchas()` function from running as it does not set the `lastElementUpdated` property when updating the submission.

divporter: yeah but this function will run nearly 10 seconds later, by which time it will have lost the race.

divporter: hmmmm but memoing....

divporter: Just tested with a 5 second delay on disabling the form and it worked.
With a 15 second delay it obviously triggered autosave. So I'm happy with this.

mymattcarroll: As discussed, im happy if you are happy.

PR #690, opened 2024-06-11
ON-41110 # add role=group to compliance element

divporter:


src/form-elements/FormElementCheckBoxes.tsx
```diff
@@ -135,6 +135,11 @@ function FormElementCheckboxes({
                 className="buttons ob-buttons ob-buttons-radio"
                 role="group"
                 aria-labelledby={`${id}-label`}
+                aria-describedby={
+                  element.hint && element.hintPosition === 'BELOW_LABEL'
+                    ? `${id}-hint`
+                    : undefined
+                }
```
jdawg093, line 142: These should be using `useElementAriaDescribedby(id, element)`

PR #700, opened 2024-07-02
ON-41508 # Added bottom collapse button to sections

Zaxist:


src/form-elements/FormElementSection.tsx
```diff
@@ -160,13 +171,36 @@ function FormElementSection<T extends FormTypes._NestedElementsElement>({
...
         }}
       >
-        <OneBlinkFormElements
-          {...props}
-          displayValidationMessages={displayValidationMessage}
-          onLookup={handleLookup}
-          elements={element.elements}
-          onUpdateFormElements={handleUpdateNestedFormElements}
-        />
+        {element.canCollapseFromBottom ? (
```
divporter, line 174: My opinion: better to create a component that wraps OneBlinkFormElements and either renders the buttons and {children} or just {children}.  I can see someone coming along later and updating the top instance of OneBlinkFormElements and not the bottom one.

PR #701, opened 2024-07-08
ON-41824 # fixed elements with options allowing empty strings

divporter:


src/components/ArcGISWebMap.tsx
```diff
@@ -26,7 +26,7 @@ function FormElementArcGISWebMap({ element, id, ...props }: Props) {
   const layerPanelRef = React.useRef<Expand | null>(null)
   const mapGalleryPanelRef = React.useRef<Expand | null>(null)
   const [isLoading, setIsLoading] = React.useState<boolean>(true)
-  const isPageVisible = useIsPageVisible()
+  const isPageVisible = usePage()
```
mymattcarroll, line 29:
```suggestion
  const { isPageVisible } = usePage()
```

src/form-elements/FormElementSignature.tsx
```diff
@@ -13,7 +13,7 @@ import AttachmentStatus from '../components/renderer/attachments/AttachmentStatu
 import { canvasToBlob } from '../services/blob-utils'
 import ImagePreviewUnavailable from '../components/renderer/attachments/ImagePreviewUnavailable'
 import { FormElementValueChangeHandler } from '../types/form'
-import useIsPageVisible from '../hooks/useIsPageVisible'
+import useIsPageVisible from '../hooks/usePage'
```
mymattcarroll, line 16: need to change this to use boolean instead of object

src/form-elements/FormElementNumber.tsx
```diff
@@ -5,7 +5,7 @@ import LookupButton from '../components/renderer/LookupButton'
 import { FormTypes } from '@oneblink/types'
 import FormElementLabelContainer from '../components/renderer/FormElementLabelContainer'
 import { FormElementValueChangeHandler, IsDirtyProps } from '../types/form'
-import useIsPageVisible from '../hooks/useIsPageVisible'
+import useIsPageVisible from '../hooks/usePage'
```
mymattcarroll, line 8: need to change this to use boolean instead of object

src/components/renderer/LookupNotification.tsx
```diff
@@ -150,9 +155,17 @@ function LookupNotificationComponent({
           }
         }
 
+        const clonedElements = _cloneDeep(definition.elements)
+
+        formElementsService.forEachFormElement(clonedElements, (el) => {
+          if (el.type === 'page' && el.id === pageId) {
+            el.elements = allElements
+          }
+        })
```
mymattcarroll, line 164: This seems very _sledge hammer to nail_ as an approach, could you please walk me through this change.

mymattcarroll: Dont we just need to remove line 95 in `src/hooks/useFormElementOptions.ts` to solve this problem?

mymattcarroll: or change the validation to not allow empty strings by setting `allowEmpty` to `false` in `src/services/form-validation.ts`.

https://validatejs.org/#validators-presence

src/components/ArcGISWebMap.tsx
divporter: derp

src/form-elements/FormElementSignature.tsx
divporter: derp yep

src/components/renderer/LookupNotification.tsx
divporter: I started with allowEmpty, but there's a bigger problem that if the elements being filled are on another page then they won't go through the generateDefaultData function and then we're allowing all sorts of incorrect values onto the submission body some of which may be caught by the validator but I would guess a lot wouldn't.

mymattcarroll: `generateDefaultData()` should not be validating the data that is applied to the submission. It should be setting default values for new elements that are being injected into the form. 

Either way, can we please create a new bug ticket to deal with this problem and get it prioritised/refined/planned? Avoiding scope creep was one of the outcomes of our last retro regarding P1s.

src/hooks/useFormElementOptions.ts
```diff
@@ -90,11 +90,7 @@ export default function useFormElementOptions<T>({
       return
     }
 
-    if (
-      typeof value === 'string' &&
-      value &&
-      !withInjectedOptions.some((option) => value === option.value)
-    ) {
+    if (!withInjectedOptions.some((option) => value === option.value)) {
```
mymattcarroll, line 93: I think we want to keep the `typeof value === 'string'` check here to avoid iterating over elements unnecessarily.

src/components/renderer/LookupNotification.tsx
divporter: Well, it's currently validating pre fill data.  Are you saying that validating pre fill is what should be split out and that is the separate story? Was this bug a P1? I'd also argue the solution isn't scope creep because the problem can only be reproduced if the elements being filled are on a different page, otherwise the validation works just fine.

PR #706, opened 2024-08-06
ON-42513 # process pending queue if status updated

divporter:


src/hooks/usePendingSubmissions.tsx
```diff
@@ -200,6 +200,9 @@ export function PendingSubmissionsContextProvider({
             processPendingQueue()
             break
           }
+          case 'EDIT_CANCELLED': {
```
mymattcarroll, line 203: as discussed, comment and move into ADDITION case

PR #859, opened 2025-06-04
ap-3322 # Added `replaceInjectablesOverrides` prop to `OneBlinkFormBase`

Zaxist:
## Requester Checklist

Please only check the items that you have actioned. Do not check items that are not applicable to your PR.

### Implementation

- [ ] Have you tested your implementation locally?
- [ ] If applicable, has an appropriate changelog entry been added? Do not include if you are fixing/changing something that is only in the current release.
- [ ] There are no warnings that have been suppressed unnecessarily
- [ ] Have automated tests been added, or have related ones been updated to cover the change?
- [ ] Have all OneBlink dependency updates been completed (eg apps / apps-react / types etc).
- [ ] Have you ensured this change does not add unwanted dependencies?
- [ ] If this PR contains a refactor, have relevant Jira testing tasks added?
- [ ] Changes that will knowingly make the feature/bug incomplete have been commented with TODO and a description of what needs to be done to finish the feature/bug
- [ ] Any changes made to public APIs have been reflected in the documentation
- [ ] Have you isolated business logic where possible to allow for unit testing?

### Logging and Debugging

- [ ] Are the error messages, if any, informative?
- [ ] Are there enough log events and are they written in a way that allows for easy debugging?
- [ ] "Debugging" code removed
- [ ] Front-end: No erroneous `Console.WriteLines`

### Readability

- [ ] All class, variable, property and method modifiers are provided with the smallest scope possible
- [ ] New files, variables and functions are descriptive/comprehensible and named consistently.
- [ ] There is no dead code (unreachable code)
- [ ] There is no usage of [magic numbers](<https://en.wikipedia.org/wiki/Magic_number_(programming)>)
- [ ] There is no commented out code.
- [ ] In hard-to-understand areas, comments exist and describe rationale or reasons for decisions in code

### Security

- [ ] All personal data inputs are checked (for the correct type, length/size, format, and range).
- [ ] No sensitive information is logged or visible in a stacktrace
- [ ] Are authorization and authentication handled correctly?
- [ ] Is (user) input validated, sanitized, and escaped to prevent security attacks such as cross-site scripting or SQL injection?
- [ ] Is data retrieved from external APIs or libraries checked for security issues?
- [ ] Do API endpoints return appropriate status codes

## Reviewer Guide

- It is important that you understand the purpose of the PR.
- You are encouraged to engage with the requestor if you do not understand any of the proposed code changes/additions/deletions.
- Do you, the reviewer, understand what the code does? Do you think a specific expert, like a security expert or a usability expert, should look over the code before it can be accepted?
- Is a framework, API, library, or service used that should not be used? Are there alternatives you could recommend?
- Are there existing hooks/components/functions in the same code base that could be utilised?
- When reviewing tests, attempt to identify missing edge cases that may be relevant to the proposed implementation
- Ensure you check for:
  - Security
  - Scalability
  - Performance
  - Maintainability

### DIFF
--- a/src/OneBlinkFormBase.tsx
+++ b/src/OneBlinkFormBase.tsx
@@ -61,6 +61,7 @@ import MaterialIcon from './components/MaterialIcon'
 import ReCAPTCHA from 'react-google-recaptcha'
 import ValidationErrorsCard from './components/ValidationErrorsCard'
 import { sendGoogleAnalyticsEvent } from './utils/sendGoogleAnalyticsEvent'
+import { useUserProfileForInjectablesOutsideContext } from './hooks/useUserProfileForInjectables'
 
 export type OneBlinkReadOnlyFormProps = {
   /**
@@ -236,7 +237,10 @@ function OneBlinkFormBase({
   replaceInjectablesOverrides,
 }: Props) {
   const isOffline = useIsOffline()
-  const { isUsingFormsKey, userProfile } = useAuth()
+  const { isUsingFormsKey } = useAuth()
+  const userProfileForInjectables = useUserProfileForInjectablesOutsideContext(
+    replaceInjectablesOverrides?.userProfile,
+  )
   const captchasRef = React.useRef<Array<ReCAPTCHA>>([])
 
   const theme = React.useMemo(
@@ -791,7 +795,7 @@ function OneBlinkFormBase({
         elements: definition.elements,
         submission: submissionData.submission,
         taskContext: taskContextValue,
-        userProfile: userProfile ?? undefined,
+        userProfile: userProfileForInjectables,
       })
       setIsPreparingToSubmit(false)
       resetRecaptchas()
@@ -818,7 +822,7 @@ function OneBlinkFormBase({
       allowNavigation,
       definition,
       taskContextValue,
-      userProfile,
+      userProfileForInjectables,
       resetRecaptchas,
       onSubmit,
       captchaSiteKey,
--- a/src/hooks/useReplaceableText.ts
+++ b/src/hooks/useReplaceableText.ts
@@ -1,22 +1,16 @@
-import { localisationService, authService } from '@oneblink/apps'
-import { useMemo, useCallback } from 'react'
+import { localisationService } from '@oneblink/apps'
+import { useMemo } from 'react'
 import useFormSubmissionModel from './useFormSubmissionModelContext'
 import { useRepeatableSetIndexText } from '../form-elements/FormElementRepeatableSet'
 import useTaskContext from './useTaskContext'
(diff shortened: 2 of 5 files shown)


src/OneBlinkFormBase.tsx
```diff
@@ -82,6 +86,11 @@ export type OneBlinkReadOnlyFormProps = {
    * scheduled task associated with a group via a form submission
    */
   taskGroupInstance?: ScheduledTasksTypes.TaskGroupInstance
+  /**
+   * Override the default behaviour when injecting values, such as in an info
+   * element or a default value.
+   */
+  replaceInjectablesOverrides?: ReplaceInjectablesOverrides
```
mymattcarroll, line 93: More react-y to make this `userProfile` so that we can re-render when the object changes. There are couple of other places we use the `userProfile` from `useAuth()`. We will need to ensure we not using `useAuth()` for any user profile data.

Zaxist: Well thats sort of the point. I don't want to replace the current logged in user everywhere. Only in the replaceable value stuff.

Zaxist: Anything else I would consider a side effect.

Zaxist: Although, it looks like the only other place using `userProfile` is the dynamic option set replaceable stuff, which does the seem thing but in its own way. so that will need to update too.

mymattcarroll: The other touch point we will need to cater for is injecting dynamic options into options based form elements. Im not sure if that is catered for already with the code change you have made so far.

mymattcarroll: Agreed.

src/hooks/useReplaceableText.ts
```diff
@@ -1,20 +1,30 @@
...
   const textWithIndex = useRepeatableSetIndexText(text)
   const { formSubmissionModel, elements } = useFormSubmissionModel()
   const { task, taskGroup, taskGroupInstance } = useTaskContext()
-  return React.useMemo(() => {
+
+  const getUserProfile = useCallback(() => {
+    if (replaceInjectablesOverrides?.getUserProfile) {
+      return replaceInjectablesOverrides.getUserProfile()
+    }
+    return authService.getUserProfile()
```
mymattcarroll, line 18: Can we swap this to user profile from `useAuth()` please? This will not re-render if user profile changes at the moment from `authService`.

src/OneBlinkFormBase.tsx
mymattcarroll: Does seem strange to require a function when we don't really have a use case for passing a function but I am ok with it if you would prefer a function.

